PR #29, opened 2022-07-16
Add GenerationMixin class

artek0chumak:
Add generation abstraction, that's using inference_session.
Added modes:
- Greedy, top-k/top-p sampling
- Multibatch generation
- Constraint abstraction
In the future, I'll add prefix-tuned generation, beam-search and more hf-like stuff.

### DIFF
--- a/src/server/backend.py
+++ b/src/server/backend.py
@@ -1,5 +1,5 @@
 """Code for serving bloom blocks via hivemind-server"""
-from typing import Sequence, Tuple
+from typing import Optional, Sequence, Tuple
 
 import torch
 from hivemind.moe.server.module_backend import ModuleBackend
--- a/src/server/backend.py
+++ b/src/server/backend.py
@@ -1,5 +1,5 @@
 """Code for serving bloom blocks via hivemind-server"""
-from typing import Optional, Sequence, Tuple
+from typing import Sequence, Tuple
 
 import torch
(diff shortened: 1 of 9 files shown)


src/server/backend.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, *args, memory_cache: MemoryCache, **kwargs):
         for name, buf in self.module.named_buffers():
             assert not buf.requires_grad, f"Bloom layer parameters must not accumulate gradients, but {name} does"
 
-        self.inference_pool = TaskPool(self.inference_step, max_batch_size=1, name=f"{self.name}_inference")
+        self.inference_pool = TaskPool(self.inference_step, max_batch_size=4096, name=f"{self.name}_inference")
```
justheuristic, line 26: This can have an adverse effect of grouping together concurrent requests for inference into one pytorch call. The current inference code will break in that case.

justheuristic (commented):
setting max batch size other than 0 in TaskPool for inference will cause it to merge requests from different users, which is not supported on the backend (both queries will fail)

There are several options to work around that:
- implement multi-source inference on the backend side
- make a custom task pool for inference
- set batch size 1 for now, fix in a subsequent PR

artek0chumak:
I fix test/black/isort. Remove change in max_batch_size in server/handler.py.

PR #30, opened 2022-07-18
Decouple make_sequence and move to RemoteSequenceManager

justheuristic:
This PR:
- [x] finish renaming RemoteSequenceInfo -> RemoteSequenceManager (why: if it was an *Info, user would expect it to be similar to a dataclass; whereas in actuality, the class is doing heavy network interactions on its own)
- [x] implement RemoteSequenceManager.make_sequence (from https://pastebin.com/uXgy2U8B )
- [x] make RemoteSequentialInferenceSession use RemoteSequenceManager.make_sequence
- [x] make tests pass again
- [x] make it possible to create inference session without RemoteTransformerBlock
- [x] make a standalone test for RemoteSequential
- [x] rollback convert-model


justheuristic:
Note: planned to implement fault tolerance in the same PR, will probably move to the next one

PR #33, opened 2022-07-21
[DRAFT] Routing for lowest inference time

justheuristic:
TODO:
- [x] RemoteSequenceInfo
- [x] RoutingStrategy interface - and forward kwargs from init
- [x] RandomRoutingStrategy
- [ ] change update_ into a background thread, implement RemoteSequenceManager.run
- [ ] move lock_changes from RemoteSequenceManager to RemoteSequenceInfo
- [ ] basic tests for RemoteSequenceManager
- [ ] make it pass tests again
- [ ] move RemoteSequenceManager.rpc_info to RemoteSequenceInfo.rpc_info (make manager.rpc_info point to the new location)
- [ ] change RandomRoutingStrategy to trigger refresh in case there is no path - as opposed to crashing in on_update
- [ ] make some warnings that indicate that the sequence is not ready
- [ ] EFSearchStrategy
- [ ] revert run-tests.html

Things changed in this PR:
- RemoteSequenceManager will now run a **background thread** that looks up DHT entries periodically
  - previously, manager would reload all DHT entries on each RemoteSequential.forward, increasing latency
- RemoteSequenceManager.make_sequence now supports multiple routing strategies

Things beyond the scope of this PR:
- __Load-balanced routing__ will be done in a subsequent PR
- __Routing code decomposition / support for user-defined strategies__ - important, but complicated and not urgent -> later
- __Optimized model init__ - prepare sequence manager in parallel with loading embeddings? save latency and blacklists on disk?
- __Cleaning up src.client file structure__ - there will come a day when we organize the client-side more neatly. But that day is not today.

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -28,7 +28,7 @@ jobs:
           pip install -r requirements.txt
       - name: Delete previous model, if exists
         run: |
-          export HF_TAG=$(python -c "import os; print(os.environ.get('GITHUB_BASE_REF') or os.environ.get('GITHUB_REF_NAME'))")
+          export HF_TAG=$(python -c "import os; print(os.environ.get('GITHUB_REF_NAME') or 'main')")
           python -c "from huggingface_hub import delete_repo; delete_repo(token='$BLOOM_TESTING_WRITE_TOKEN', \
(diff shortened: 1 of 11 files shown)


justheuristic:
Status: this PR is hopelessly out of sync with master. It will be re-written from scratch..

PR #34, opened 2022-07-22
Rebalance swarm when necessary

borzunov:
**Status:** I've tested it on simple cases, it works.

**Future work:**

- [ ] Sometimes server fails to restart due to shmem errors. Rebalancing is disabled by default until we solve that.
- [ ] `Server` should follow `ModuleContainer` status and restart it if it crashes.
- [ ] I'd like to add some docs on properties of the greedy algorithm in `block_selection.py`.
- [ ] Add unit tests for algorithms from `block_selection.py`.
- [ ] (maybe) Add functional tests for real server rebalancing.

### DIFF
--- a/cli/run_server.py
+++ b/cli/run_server.py
@@ -79,6 +79,13 @@ def main():
     parser.add_argument('--custom_module_path', type=str, required=False,
                         help='Path of a file with custom nn.modules, wrapped into special decorator')
     parser.add_argument('--identity_path', type=str, required=False, help='Path to identity file to be used in P2P')
+
+    parser.add_argument("--min_balance_quality", type=float, default=0.0,
+                        help="Rebalance the swarm if its balance quality (a number in [0.0, 1.0]) "
+                             "goes below this threshold. Default: rebalancing is disabled")
+    parser.add_argument("--mean_balance_check_period", type=float, default=150,
+                        help="Check the swarm's balance every N seconds (and rebalance it if necessary)")
+
     parser.add_argument("--use_auth_token", type=str, default=None, help="auth token for from_pretrained")
     parser.add_argument('--load_in_8bit', action='store_true', help='Convert the loaded model into mixed-8bit quantized model.')
 
--- a/cli/run_server.py
+++ b/cli/run_server.py
@@ -43,7 +43,7 @@ def main():
                         help='Use this many threads to pass results/exceptions from Runtime to Pools')
     parser.add_argument('--inference_max_length', type=int, default=16384,
                         help='Maximum total sequence length permitted per inference, defaults to 16384 tokens')
-    parser.add_argument('--cache_dir', type=str, default=None, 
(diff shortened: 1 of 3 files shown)


src/server/server.py
```diff
@@ -29,76 +30,13 @@
 
 
 class Server(threading.Thread):
-    """Serves one or more bloom layers for inference, forward and backward; announces oneself to the DHT"""
+    """
+    Runs ModuleContainer, periodically checks that the network is balanced,
```
justheuristic, line 34: [nit] ModuleContainer sounds somewhat too generic for a class that announces modules to DHT, accepts requests from p2pd, etc

Maybe
* Server -> LoadBalancedServer 
* ModuleContainer -> Server
?

src/server/block_selection.py
```diff
@@ -1,18 +1,101 @@
...
+import numpy as np
+from hivemind import PeerID, get_logger
 
 from src.data_structures import RemoteModuleInfo, ServerState
 
+logger = get_logger(__file__)
+
+
+@dataclass
+class Span:
```
justheuristic, line 13: [nit] please exclude this from __all__

src/server/server.py
```diff
@@ -169,16 +135,174 @@ def create(
...
+
+    def shutdown(self):
+        self.stop.set()
+
+        self.dht.shutdown()
+        self.dht.join()
+
+
+class ModuleContainer(threading.Thread):
+    """Serves a set of specific Bloom layers for inference, forward, and backward. Announces itself over the DHT."""
```
justheuristic, line 216: ServedLayers?

justheuristic (approved):

src/server/server.py
```diff
@@ -121,46 +59,74 @@ def create(
         expiration: Optional[float] = None,
         prefetch_batches: int = 1,
         sender_threads: int = 1,
-        max_block_selection_delay: float = 1,
+        mean_block_selection_delay: float = 0.5,
+        mean_balance_check_period: float = 150,
+        min_balance_quality: float = 0.8,
```
borzunov, line 64:
```suggestion
        min_balance_quality: float = 0.0,
```

TODO: Disable rebalancing by default, unless we solve issues with shmem

cli/run_server.py
```diff
@@ -104,7 +104,7 @@ def main():
     use_auth_token = args.pop("use_auth_token")
     args["use_auth_token"] = True if use_auth_token in ("True", "true", "") else use_auth_token
 
-    server = Server.create(**args, start=True, compression=compression, attn_cache_size=attn_cache_size)
+    server = Server(**args, compression=compression, attn_cache_size=attn_cache_size, start=True)
```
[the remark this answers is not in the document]
borzunov: [discussed during the call]

src/server/server.py
borzunov: I agree that the name is not perfect but I didn't come up with better options. If we find one, we can rename this class later.

Issue #16, opened 2022-07-01
Client-side code improvements [Yozh-todo-list]


justheuristic:
### Important stuff

- [ ] optimal (instead of random) routing in RemoteSequenceManager
- [x] fault tolerance in RemoteSequentialInferenceSession
- [x] quick slicing with RemoteSequential[5:15] ( @TimDettmers )
- [x] BloomForCausalLM.generate ( @artek0chumak )
- [x] remove lm_head WEIGHT from checkpoint and/or model, convert BloomModel instead (to save 2x client RAM) [@dbaranchuk did it]

### Minor stuff
This list contains all the things i've delegated to future me. They are typically too obscure for anyone else to worry about

- [x] fault tolerance inside RemoteTransformerBlock (as opposed to RemoteSequential)

- [x] in SequenceInfo, keep block_infos as TimedStorage instead of just RemoteSequenceInfo for automatic removal of expired servers
- [x] make RemoteSequential print the state of the model ( @TimDettmers )
- [x] less frequent DHT lookups in RemoteSequenceInfo
- [x] change BloomForCausalLM.forward to call RemoteSequential instead of working layer-by-layer (why: to allow chaining subsequent blocks efficiently)

Note from #45  by @borzunov 

Please consider running a loop instead, maybe using just --num_blocks without explicit --block_indices. If it gets more complicated than just repeating something N times, please move it to a Python script.

justheuristic:
As of 2022.01.02, everything except smart routing has been implemented. Smart routing is currently in progress (see #12 ).

PR #45, opened 2022-08-17
Reduce vocabulary size in test model, fix bug in routing when overlapped

justheuristic:
More than half of the test 560m model parameters is it's embedding layer (250_880 x 1024).
This adds 1-2 GB of RAM depending on the test scenario
This PR reduces this vocabulary size to save memory during conversion, keeping only the first 50k tokens

As a result,
* tests that load client-side embeddings need significantly less RAM
* we can now run CI tests with 4 servers instead of 2 - needed to test routing - see bugs uncovered
* some of the servers now use load balancing
* CI convert_model now takes 4-5 minutes (was 6-7)

Running CI with 4 servers uncovered a bug:
https://github.com/bigscience-workshop/petals/runs/7879399020?check_suite_focus=true
which is now also fixed

This PR also temporarily pins an older bnb version, pending @TimDettmers 's fix for cpuonly

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -36,8 +36,8 @@ jobs:
           export HF_TAG=$(python -c "import os; print(os.environ.get('GITHUB_HEAD_REF') or os.environ.get('GITHUB_REF_NAME'))")
           python -m cli.convert_model --model bigscience/bloom-560m  --output_path ./converted_model \
             --output_repo bloom-testing/test-bloomd-560m-$HF_TAG --use_auth_token $BLOOM_TESTING_WRITE_TOKEN \
-            --resize_token_embeddings 10_000  # reduce embeddings size to save memory
-
+            --resize_token_embeddings 10000
+          
 
   run-tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -26,12 +26,10 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements.txt
-      - name: Delete test models older than 72 hours
+      - name: Delete any test models older than 72 hours
         run: |
-          export HF_TAG=$(python -c "import os; print(os.environ.get('GITHUB_HEAD_REF') or os.environ.get('GITHUB_REF_NAME'))")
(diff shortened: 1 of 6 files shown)


.github/workflows/run-tests.yaml
```diff
@@ -26,6 +26,11 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements.txt
+      - name: Delete test models older than 72 hours
+        run: |
+          export HF_TAG=$(python -c "import os; print(os.environ.get('GITHUB_HEAD_REF') or os.environ.get('GITHUB_REF_NAME'))")
+          python -c "from huggingface_hub import delete_repo; delete_repo(token='$BLOOM_TESTING_WRITE_TOKEN', \
+          name='test-bloomd-560m-$HF_TAG', organization='bloom-testing')" || true
```
borzunov, line 33: Please consider moving that to a small Python script in the repo. Having a Bash command running Python, whose results are passed to another Bash command running Python, is not good.

borzunov, line 33: Please consider moving that to a small Python script in the repo. Having a Bash command running Python, whose results are passed to another Bash command running Python, is not good.

.github/workflows/run-tests.yaml
```diff
@@ -76,21 +82,41 @@ jobs:
...
           
-          python -m cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 12:24 \
+          python -m cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 12:22 \
             --torch_dtype float32 --initial_peers $INITIAL_PEERS --throughput 1 &> server2.log &
           SERVER2_PID=$!
 
-          sleep 60  # wait for server to download layers
+          sleep 10 # wait for initial servers to declare blocks, then let server decide which blocks to serve
+
+          python -m cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 0:6 \
```
borzunov, line 100: Please consider running a loop instead, maybe using just `--num_blocks` without explicit `--block_indices`. If it gets more complicated than just repeating something N times, please move it to a Python script.

.github/workflows/run-tests.yaml
justheuristic: done, see tests/srcipts/remove_old_models.py

.github/workflows/run-tests.yaml
justheuristic: moved to #16

PR #50, opened 2022-08-28
Fix recovering for sequential_backward

dbaranchuk:


justheuristic (approved):
Got it, we need some automatic test for failed backward - please create an issue about it

PR #47, opened 2022-08-21
Priority tasks

GreenFatGuy:
Here we implement first version of "financial" side of the cooperation in petals
TODO:
- [x] Tests

PR #52, opened 2022-08-28
Inference: require max sequence length instead of assuming 2048

justheuristic:
- Maximum length is now provided in `.inference_session(max_length=100)`
   - previously, we would always assume max length = 2048

- added a generic way to forward **kwargs to inference session
  - for compatibility with #47 
  - Note to @borzunov : it does *not* pass them arbitrarily, but instead checks for kwarg names at the bottom level
- run_server can be started with a custom max_length for inference
- renamed --cache_size_bytes to --attention_cache_bytes (to avoid collision with --cache_dir)
- --attn_cache_bytes can now support humane file sizes (e.g. 300MB instead of 314572800)
- made some server-side errors more human-readable to user (e.g. when max length is exceeded)

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -81,7 +81,7 @@ jobs:
 
           python -m cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 0:12 \
             --torch_dtype float32 --identity tests/test.id --host_maddrs /ip4/127.0.0.1/tcp/31337 \
-            --throughput 1 --attention_cache_size 0.2GiB &> server1.log &
+            --throughput 1 --attn_cache_size 0.2GiB &> server1.log &
           SERVER1_PID=$!
           
           sleep 5  # wait for the first server to initialize DHT
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -81,7 +81,7 @@ jobs:
 
           python -m cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 0:12 \
             --torch_dtype float32 --identity tests/test.id --host_maddrs /ip4/127.0.0.1/tcp/31337 \
-            --throughput 1 --attention_cache_bytes 0.2GiB &> server1.log &
+            --throughput 1 --attention_cache_size 0.2GiB &> server1.log &
           SERVER1_PID=$!
           
           sleep 5  # wait for the first server to initialize DHT
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -81,7 +81,7 @@ jobs:
 
           python -m cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 0:12 \
             --torch_dtype float32 --identity tests/test.id --host_maddrs /ip4/127.0.0.1/tcp/31337 \
-            --throughput 1 &> server1.log &
+            --throughput 1 --attention_cache_bytes 0.2GiB &> server1.log &
           SERVER1_PID=$!
           
           sleep 5  # wait for the first server to initialize DHT
--- a/cli/run_server.py
+++ b/cli/run_server.py
@@ -34,8 +34,8 @@ def main():
                         help='Minimum required batch size for all expert operations')
     parser.add_argument('--max_batch_size', type=int, default=16384,
                         help='The total number of tokens in the same batch will not exceed this value')
-    parser.add_argument('--inference_max_length', type=int, default=None,
-                        help='Maximum total sequence length permitted per inference, defaults to max_batch_size tokens')
+    parser.add_argument('--inference_max_length', type=int, default=16384,
+                        help='Maximum total sequence length permitted per inference, defaults to 16384 tokens')
     parser.add_argument('--cache_dir', type=str, default=None, 
                         help='Path to a directory in which a downloaded pretrained model configuration should be cached if the standard cache should not be used.')
     parser.add_argument('--device', type=str, default=None, required=False,
--- a/cli/run_server.py
+++ b/cli/run_server.py
@@ -107,7 +107,7 @@ def main():
     use_auth_token = args.pop("use_auth_token")
     args["use_auth_token"] = True if use_auth_token in ("True", "true", "") else use_auth_token
 
-    server = Server.create(**args, start=True, compression=compression, cache_size_bytes=attention_cache_bytes)
+    server = Server.create(**args, start=True, compression=compression, attention_cache_bytes=attention_cache_bytes)
 
     try:
         server.join()
--- a/cli/run_server.py
+++ b/cli/run_server.py
@@ -101,7 +101,8 @@ def main():
     attn_cache_size = args.pop("attn_cache_size")
     if attn_cache_size is not None:
         attn_cache_size = parse_size_as_bytes(attn_cache_size)
-    assert isinstance(attn_cache_size, (int, type(None))
+    assert isinstance(
+        attn_cache_size, (int, type(None))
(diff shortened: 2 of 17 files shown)


cli/run_server.py
```diff
@@ -8,6 +8,20 @@
 use_hivemind_log_handler("in_root_logger")
 logger = get_logger(__file__)
 
+import re
+
+
+def parse_size_as_bytes(size: str) -> int:
+    """parse human-readable data size e.g. 1.5GB, based on https://stackoverflow.com/a/42865957/2002471"""
```
justheuristic, line 15: This cuts some corners (e.g. 1GB = 1GiB). We can get it right all the time by using https://pypi.org/project/humanfriendly/ , but I'm not sure it justifies the extra depenedcy

@borzunov any preferences?

cli/run_server.py
```diff
@@ -8,6 +8,20 @@
 use_hivemind_log_handler("in_root_logger")
 logger = get_logger(__file__)
 
+import re
```
borzunov, line 11: Let's move this to the top of the file

cli/run_server.py
```diff
@@ -32,16 +46,19 @@ def main():
...
     parser.add_argument('--cache_dir', type=str, default=None, 
                         help='Path to a directory in which a downloaded pretrained model configuration should be cached if the standard cache should not be used.')
-    parser.add_argument('--cache_size_bytes', type=int, default=None,
-                        help='The size of memory cache for storing past attention keys/values between inference steps')
     parser.add_argument('--device', type=str, default=None, required=False,
                         help='all experts will use this device in torch notation; default: cuda if available else cpu')
     parser.add_argument("--torch_dtype", type=str, default="auto",
                         help="Use this dtype to store block weights and do computations. "
                              "By default, respect the dtypes in the pre-trained state dict.")
+    parser.add_argument('--attention_cache_bytes', type=str, default=None,
```
borzunov, line 59:
```suggestion
    parser.add_argument('--attn_cache_size', type=str, default=None,
```

I strongly advise to replace `cache_bytes` to `cache_size` because:

- Specifying the size in bytes is standard across Python libs
- Moreover, here the size can be specified in any units

Also, I weakly advise to replace `attention` to `attn` because it's much shorter (but still understandable to everyone). This is optional though.

tests/test_chained_calls.py
```diff
@@ -63,7 +63,7 @@ def test_chained_inference_exact_match(atol_inference=1e-4):
     inputs = torch.randn(1, 8, config.hidden_size)
 
     outputs_inference = []
-    with remote_block.inference_session() as sess:
+    with remote_block.inference_session(max_length=inputs.shape[1]) as sess:
```
borzunov, line 66: Maybe add a test checking what happens if we exceed `max_length`?

src/server/server.py
```diff
@@ -135,13 +138,15 @@ def create(
         assert (block_indices is None) != (num_blocks is None), "please specify num_blocks or block_indices, not both"
         if expiration is None:
             expiration = max(2 * update_period, MAX_DHT_TIME_DISCREPANCY_SECONDS)
+        if inference_max_length is None:
+            inference_max_length = max_batch_size
```
borzunov, line 142: Why do we assign max batch size to max sequence size?

cli/run_server.py
```diff
@@ -81,10 +98,17 @@ def main():
     compression_type = args.pop("compression")
     compression = getattr(CompressionType, compression_type)
 
+    attention_cache_bytes = args.pop("attention_cache_bytes")
+    if attention_cache_bytes is not None:
+        attention_cache_bytes = parse_size_as_bytes(attention_cache_bytes)
+    assert isinstance(attention_cache_bytes, (int, type(None))), (
+        "unrecognized value for attention_cache_bytes," " examples: 1.5GB or 1500MB or 1572864000"
```
borzunov, line 105:
```suggestion
        "unrecognized value for attention_cache_bytes, examples: 1.5GB or 1500MB or 1572864000"
```

src/server/handler.py
```diff
@@ -43,7 +44,15 @@ async def rpc_inference(
...
             request = await anext(requests)
             requested_uids = self._check_uids(request.uid)
+            metadata = MSGPackSerializer.loads(request.metadata) if request.metadata else {}
             requested_backends = tuple(self.module_backends[uid] for uid in requested_uids)
+            max_length = metadata.get("max_length")
+
+            if not requested_uids:
+                raise ValueError("User must specify at least one block for inference, but got none")
+            assert isinstance(max_length, int), f"rpc_inference metadata must contain int seq_length, got {max_length}"
+            if max_length not in range(0, self.inference_max_length):
```
borzunov, line 54:
```suggestion
            if not 0 <= max_length <= self.inference_max_length:
```

* `max_length` should be allowed to be equal to `inference_max_length` (currently it's only allowed to be `inference_max_length - 1`)
* Double inequality is a bit better than `range()`, so there's no questions from people who are confused by Python 2 (`range()` generated the whole array there, so this line would have caused OOM)

cli/run_server.py
borzunov: * I'm okay with adding this light dependency.
* I'm not okay (weakly) with the same behavior for `GB` and `GiB`.

To support both `GB` and `GiB` in a correct way, you can actually simplify the code to do smth like:

```python
units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB']
unit = unit.upper()
if 'I' in unit:
    unit = 1024 ** units.index(unit.replace('I', ''))
else:
    unit = 1000 ** units.index(unit)
```

In other words, you can use the same array of units, just with different bases :)

borzunov (changes requested):

cli/run_server.py
justheuristic: [done both]

src/server/server.py
justheuristic: both are (meant to be) in tokens

Would you prefer to set it to a constant by default?

cli/run_server.py
justheuristic: [merged]

tests/test_chained_calls.py
justheuristic: Done, see test_block_exact_match.py

.github/workflows/run-tests.yaml
```diff
@@ -81,7 +81,7 @@ jobs:
 
           python -m cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 0:12 \
             --torch_dtype float32 --identity tests/test.id --host_maddrs /ip4/127.0.0.1/tcp/31337 \
-            --throughput 1 &> server1.log &
+            --throughput 1 --attention_cache_size 0.2GiB &> server1.log &
```
borzunov, line 84:
```suggestion
            --throughput 1 --attn_cache_size 0.2GiB &> server1.log &
```

src/server/handler.py
```diff
@@ -43,7 +44,15 @@ async def rpc_inference(
             print("OPENED RPC_INFERENCE")
             request = await anext(requests)
             requested_uids = self._check_uids(request.uid)
+            metadata = MSGPackSerializer.loads(request.metadata) if request.metadata else {}
             requested_backends = tuple(self.module_backends[uid] for uid in requested_uids)
+            max_length = metadata.get("max_length")
+
+            if not requested_uids:
+                raise ValueError("User must specify at least one block for inference, but got none")
+            assert isinstance(max_length, int), f"rpc_inference metadata must contain int seq_length, got {max_length}"
```
borzunov, line 53:
```suggestion
            assert isinstance(max_length, int), f"rpc_inference metadata must contain int max_length, got {max_length}"
```

borzunov (changes requested):

.github/workflows/run-tests.yaml
justheuristic: done

PR #53, opened 2022-08-29
Make attention cache wait until memory is freed

justheuristic:
Previously, attempting to allocate with MemoryCache that does not have enough space would throw AllocationFailed.

PR changes this behavior to the following:
- by default, wait until memory is freed by other tenants (FIFO)
- if could not allocate within timeout, throw AllocationFailed
- if allocated size is too big to fit even in empty cache, throw AllocationFailed

- [x] passes existing tests
- [x] passes manual load tests
- [ ] expose request timeout to the user - moved to the next PR

p.s. if anyone wondered: using mp.Condition will not make the code simpler, their lock behavior is slightly different to what we need here


src/server/cache.py
```diff
@@ -67,19 +71,17 @@ async def allocate_cache(self, descr: TensorDescriptor) -> AsyncContextManager[H
         assert descr.device is None and descr
         allocated_handle = None
         allocated_size_bytes = descr.numel() * torch.finfo(descr.dtype).bits // 8
+        loop = asyncio.get_event_loop()
         try:
-            async with hivemind.utils.enter_asynchronously(self.lock_metadata):
+            async with hivemind.utils.enter_asynchronously(self._lock_acquire_memory):
```
borzunov, line 76: **Important discussion:**

Note that due to this lock, only one `_wait_until_available()` can run at a time.

What if someone tries to allocate `max_size_bytes`? Are we sure that all the previous allocations will be eventually freed? If it's not the case, such a query will lead to a deadlock (`MemoryCache` won't allocate neither this large chunks, nor any future smaller chunks).

Even if it's the case, the memory will be underutilized.

Alternatively, we could run `_wait_until_available()` for all the allocations simultaneously - this way, smaller allocations would utilize the recently freed memory. However, if we have a constant stream of small and large mallocs, it may never allow large chunks to be allocated.

src/server/cache.py
```diff
@@ -67,19 +71,17 @@ async def allocate_cache(self, descr: TensorDescriptor) -> AsyncContextManager[H
...
-                        f"Could not allocate {allocated_size_bytes} bytes in cache; cache size = "
-                        f"{self.max_size_bytes} bytes; {self.current_size_bytes} already allocated."
-                    )
-
-                allocated_handle = int(self.handle_counter)
-                self.current_size_bytes += allocated_size_bytes
-                self.handle_counter += 1  # note: this will eventually overflow and it is okay
-                self._pending_messages.value += 1
-                self._pipe_send.send((allocated_handle, descr))
+                    await loop.run_in_executor(None, self._wait_until_available, allocated_size_bytes)
```
borzunov, line 78: **Important:**

Suppose the event loop has N workers in its executor. N is usually not very big.

Then, if M processes come to line 76 roughly at the same time (M >= N), this line will deadlock since no free executors are left.

src/server/cache.py
```diff
@@ -67,19 +71,17 @@ async def allocate_cache(self, descr: TensorDescriptor) -> AsyncContextManager[H
...
+                async with hivemind.utils.enter_asynchronously(self.lock_metadata):
```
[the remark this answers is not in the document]
borzunov: **Important:**

Suppose the event loop has N workers in its executor. N is usually not very big.

Then, if M processes come to line 76 roughly at the same time (M >= N), this line will deadlock since no free executors are left.

borzunov (changes requested):

src/server/cache.py
justheuristic: > Are we sure that all the previous allocations will be eventually freed?
All allocations correspond to currently running rpc_inference queries and leave no data behind.

borzunov: Okay, so the memory is deliberately allocated in FIFO mode.

Good.

src/server/cache.py
justheuristic: Fixed in hivemind

src/server/cache.py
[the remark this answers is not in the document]
justheuristic: fixed in hivemind

src/server/cache.py
```diff
@@ -88,6 +90,20 @@ async def allocate_cache(self, descr: TensorDescriptor) -> AsyncContextManager[H
                     self._pending_messages.value += 1
                     self._pipe_send.send((allocated_handle, None))  # signal runtime to free that handle
                     self.current_size_bytes -= allocated_size_bytes
+                self._memory_freed_event.set()
+
+    def _wait_until_available(self, allocated_size_bytes: int, timeout: Optional[float] = None):
+        # note: this function should only be called inside lock_metadata!
```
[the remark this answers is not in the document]
justheuristic: fixed

PR #83, opened 2022-11-01
Fix "Too many open files" during rebalancing

borzunov:
Now, the number of open files stays the same after every rebalancing.


src/server/task_pool.py
```diff
@@ -89,10 +91,14 @@ def start(self):
         self._prioritizer_thread.start()
         super().start()
 
-    def shutdown(self, timeout: Optional[float] = None):
-        self.submitted_tasks.put(None)
-        self.terminate()
```
borzunov, line 94: Terminating the process here the hard way leaves open files. I replaced it with a graceful termination.

PR #89, opened 2022-11-24
Fix floating point issues in block_selection.py

borzunov:


src/server/block_selection.py
```diff
@@ -47,17 +50,14 @@ def _compute_spans(module_infos: List[Optional[RemoteModuleInfo]]) -> Tuple[Dict
     return spans, throughputs
 
 
-def _choose_best_start(throughputs: np.ndarray, num_blocks: int, cur_start: Optional[int]) -> int:
-    options = (
-        (sorted(throughputs[i : i + num_blocks]), i != cur_start, i)
-        for i in range(0, len(throughputs) - num_blocks + 1)
-    )
+def _choose_best_start(throughputs: np.ndarray, num_blocks: int) -> int:
```
borzunov, line 53: It's not necessary to explicitly prefer the previous server position anymore. See the diff below to understand why.

justheuristic (approved):

PR #91, opened 2022-11-26
Make inference, forward, and backward fully fault-tolerant

borzunov:
In this PR:

1. **The inference session is made fully fault-tolerant.** If a server fails (or doesn't respond within timeout), it will be replaced by one (or more) servers, and the lost attention caches will be regenerated. See the screenshot of how it works below.

    Here, the 24-layer model is spread across 3 servers, then the intermediate one (holding blocks 8-14) leaves, then a new large server joins and decides to host layers 6-24 (thus closing the gap). The inference session is able to recover the span 8-14 through the new server and successfully continues inference. The results are identical to the ones without failures:

    Screenshot 2022-11-26 at 08 10 49

2. **Forward and backward are made "gap-tolerant".** If some servers leave and a gap arises in the swarm (i.e., some blocks are not handled by anyone), the forward and backward pass will not fail anymore. Instead, they will keep retrying until a new path through the model is available. I have checked that the activations/gradients calculated after failure recovery are equal to the ones calculated without failures.

3. **Fixed an important bug in forward/backward.** Sometimes, `make_sequence()` returns a sequence that is longer than requested (since the last server hosts blocks further than `end_index`). Before this PR, the code actually run the inputs through these extra blocks leading to incorrect results. This affected partial forward passes and partial/full backward passes (since they re-run partial forward passes in case of failures).

4. minor: **Renamed classes:**

    - `RemoteTransformerBlockInferenceSession ` -> `_ServerInferenceSession` (since **(a)** it actually handles the whole span instead of a single block and **(b)** it is not designed to be used outside Petals internals)
    - `RemoteSequentialInferenceSession` -> `InferenceSession` (since the previous name is too long and hardly readable)

5. While working on this PR, I have also discovered a bug in hivemind that leads to an error in Petals. As soon as it is fixed, we should update `requirements.txt` to point to a new hivemind version. See https://github.com/learning-at-home/hivemind/pull/521

### DIFF
--- a/src/client/inference_session.py
+++ b/src/client/inference_session.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
+import itertools
+import logging
+import time
 from typing import AsyncIterator, List, Optional
 
 import torch
@@ -18,6 +20,7 @@ from hivemind import (
 from hivemind.moe.client.remote_expert_worker import RemoteExpertWorker
 from hivemind.p2p import StubBase
 from hivemind.proto import runtime_pb2
+from hivemind.utils.asyncio import aiter_with_timeout
 
 from src.client.sequence_manager import RemoteSequenceManager
 from src.data_structures import CHAIN_DELIMITER, ModuleUID, RemoteSpanInfo, RPCInfo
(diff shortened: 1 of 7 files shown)


src/client/inference_session.py
```diff
@@ -42,32 +42,35 @@ def __init__(
...
         outputs_aiter: AsyncIterator,
         *,
+        timeout: float,
         max_length: int,
         points: int = 0,
     ):
         self.uid, self.rpc_info = uid, rpc_info
         self.num_blocks = uid.count(CHAIN_DELIMITER) + 1
-        # warning: this code manages async objects that are only usable inside RemoteExpertWorker's background thread;
-        # using them in any other EventLoop may cause side-effects including, headaches, diarrhea, and loss of sleep
```
borzunov, line 51: This message is now sent by having `_` at the start of the class name.

src/client/inference_session.py
```diff
@@ -154,60 +157,145 @@ def __exit__(self, *exc_details):
...
+                    block_idx = span.end
+                    break
+                except Exception as e:
+                    delay = self._sequence_manager.min_backoff * 2**attempt_no
+                    logger.warning(
+                        f"Caught exception when running inference from block {block_idx} "
+                        f"(retry in {delay:.0f} sec): {repr(e)}"
+                    )
+                    logger.debug("See detailed traceback below:", exc_info=True)
+                    time.sleep(delay)
```
justheuristic, line 286: I'd recommend that (at least) the first retry happen **without delay**

Here, if attempt_no==0, delay==min_backoff

src/client/inference_session.py
```diff
@@ -154,60 +157,145 @@ def __exit__(self, *exc_details):
...
+                try:
+                    if attempt_no >= 1:
+                        self._sequence_manager.update_()
+                        if server_idx < len(self._chosen_spans):
+                            recovery_end = self._chosen_spans[server_idx].end
+                        else:
+                            recovery_end = len(self._sequence_manager)
+                        if attempt_no == 1:
+                            logger.info(
+                                f"Entering recovery mode, remote attention caches "
```
justheuristic, line 235: [nit] "entering recovery mode" reads as "from now on, all subsequent actions will happen in this new mode". Maybe "One of remote servers failed, recovering..." ?

justheuristic (approved):
Great work!

PR #99, opened 2022-11-29
Fix dtypes in backend schemas

borzunov:
Currently, the schemas use `torch.float32`, so all inputs and outputs converted to float32 before sending and after receiving on both servers and clients. This creates a huge slowdown for the system.

This PR makes the schemas use the server's `--torch_dtype` argument (default is `torch.bloat16` for BLOOM-176B).

_added on 2022.11.29: this PR also adds

* an option for client to request a specific output compression. Use case 1: client sends quantized inputs and expects quantized inputs in return. Use case 2: client uses quantization for gradients w.r.t. activations, but keeps grads w.r.t. __prompts__ as is for greater precision.
* a comment explaining the purpose of NoSpendingPolicy - since we likely won't have it for the workshop
* a test with custom compression (janky implementation for testing purposes)


justheuristic (approved):

src/server/backend.py
```diff
@@ -37,7 +37,9 @@ def __init__(self, *args, memory_cache: MemoryCache, backend_dtype: Optional[tor
         self.backward_pool = PrioritizedTaskPool(
             self.backward, max_batch_size=max_batch_size, name=f"{self.name}_backward"
         )
-        self.dtype = backend_dtype if backend_dtype else self.module.input_layernorm.weight.dtype
```
borzunov, line 40: The default backend dtype is now chosen in `ModuleContainer.create()`, since it is also needed for the schemas.

tests/test_remote_sequential.py
```diff
@@ -4,7 +4,7 @@
 from test_utils import *
 
 from petals.bloom.from_pretrained import load_pretrained_block
-from petals.client import RemoteSequential
+from petals.client import NoSpendingPolicy, RemoteSequential
```
borzunov, line 7: Unused import?

PR #100, opened 2022-11-29
[do not merge] diff compression

justheuristic:
This PR is intended as a backup if we experience any problems with finetuning in the current branch.

TL;DR previously we tested that finetuning works with compression as in this file. However, it complicates the interface and makes the code harder to follow. It also forces us to change some _exact_match tests to approximate matches and/or complicate those tests.

There is a chance that we will be able to train p-tuning normally by casting prompt gradients to higher precision. If not, we'll have to apply this PR (3 last commints)


justheuristic:
If noone objects, this will be closed in 24 hours

PR #109, opened 2022-12-01
Update beam search algorithm with vectorized version

artek0chumak:


src/petals/utils/generation_algorithms.py
```diff
@@ -80,42 +80,44 @@ def __init__(self, num_beams: int, batch_size: int) -> None:
...
+            permuted_indexes = torch.cat(
+                [torch.arange(0, self.num_beams) * self.batch_size + i for i in range(self.batch_size)], dim=0
+            )
+            probs = probs[:, : self.num_beams][permuted_indexes]
+            probs = probs.view(self.batch_size, self.num_beams, self.num_beams)
+            self._batch_beams = self._batch_beams[:, :, None] + probs
+            self._batch_beams = self._batch_beams.view(self.batch_size, -1)
+            sorted_batch_beams, sorted_hypo_ids = torch.sort(self._batch_beams, descending=True, dim=-1)
+            self._batch_beams = sorted_batch_beams[:, : self.num_beams]
+            hypo_ids = sorted_hypo_ids[:, : self.num_beams]
```
mryab, line 100: Perhaps we should replace this with something like `torch.topk(self._batch_beams, self.num_beams)`?

src/petals/utils/generation_algorithms.py
```diff
@@ -80,42 +80,44 @@ def __init__(self, num_beams: int, batch_size: int) -> None:
...
-                            (new_beam[0] + probs[probs_idx, hypo_idx].item(), beam_idx * self.num_beams + hypo_idx)
-                        )
-                self._batch_beams[batch_idx] = sorted(new_beams, reverse=True)[: self.num_beams]
+        hypo_ids = None
+        if self._cur_num_beams > 1:
+            permuted_indexes = torch.cat(
+                [torch.arange(0, self.num_beams) * self.batch_size + i for i in range(self.batch_size)], dim=0
+            )
+            probs = probs[:, : self.num_beams][permuted_indexes]
+            probs = probs.view(self.batch_size, self.num_beams, self.num_beams)
```
mryab, line 95: I'm having some difficulties understanding the code. Why do we need to materialize a matrix (self.batch_size, self.num_beams, self.num_beams)?

artek0chumak: For each old hypothesis, we have continuations. The number of possible continuations is num_beams. We add the probability of them and select the best one after. This shape of the matrix helps with the addition of broadcasted matrices.

PR #113, opened 2022-12-02
Support --load_in_8bit on pre-Turing GPUs

justheuristic:
- Linear8bitLt now supports for pre-turing GPUs by temporarily upcasting quantized weights.
- added a test for linear8bitlt accuracy with the new fallback, the accuracy is similar than the real thing, (slightly better due to non-quantized A)
- performance is roughly halfway between the default mode and memory_efficient_backward

Alternatives considered:
- cupy - slow, casting to float internally
- triton - fast but unstable af. every 3rd attempt to matmul is a segfault
- bnb.functional.igemm (no lt) - "CuBLAS Error 8" on old GPUs

### DIFF
--- a/src/petals/utils/linear8bitlt_patch.py
+++ b/src/petals/utils/linear8bitlt_patch.py
@@ -236,6 +236,7 @@ class CustomMatMul8bitLt(MatMul8bitLt):
             C32A, SA = F.transform(CA, "col32")
             out32, Sout32 = F.igemmlt(C32A, state.CxB, SA, state.SB)
             if bias is None or bias.dtype == torch.float16:
+                # we apply the fused bias here
                 output = F.mm_dequant(out32, Sout32, SCA, state.SCB, bias=bias)
                 output = output.to(A.dtype)
             else:  # apply bias separately
@@ -250,8 +251,6 @@ class CustomMatMul8bitLt(MatMul8bitLt):
             output = output.mul_(state.SCB.unsqueeze(0).mul(1.0 / 127.0))
             if bias is not None:
                 output = output.add_(bias)
-
-        # we apply the fused bias here
 
         # 4. Mixed-precision decomposition matmul
         if coo_tensorA is not None and subA is not None:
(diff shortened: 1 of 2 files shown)


justheuristic:
Please note: the tests ran twice because this PR used to be force-pushed. This is not a bug, it will not replicate in subsequent PRs.

src/petals/utils/linear8bitlt_patch.py
```diff
@@ -123,9 +125,156 @@ def custom_matmul8bitlt(
 
 
 class CustomMatMul8bitLt(MatMul8bitLt):
-    # forward is the same as in inference-only CxB
+    # forward is the same, but we added the fallback for pre-turing GPUs
     # backward is mostly the same, but adds one extra clause (see "elif state.CxB is not None")
 
+    @staticmethod
+    def forward(ctx, A, B, out=None, bias=None, state=CustomMatmulLtState):
```
justheuristic, line 132: For simplicity: the new code in forward is limited to

- L133 - and a few times using_igemmlt was added to if-statemets
- L245-252

borzunov: For the reference, the original is [here](https://github.com/TimDettmers/bitsandbytes/blob/439f2b0c10abd3e9aade386d92810b074c69e9ec/bitsandbytes/autograd/_functions.py#L204).

src/petals/utils/linear8bitlt_patch.py
```diff
@@ -123,9 +125,156 @@ def custom_matmul8bitlt(
...
+        else:
+            A_wo_outliers = A.clone()
+            if state.idx is not None:
+                A_wo_outliers[:, state.idx.long()] = 0
+            output = torch.nn.functional.linear(A_wo_outliers, state.CB.to(A.dtype))
+            output = output.mul_(state.SCB.unsqueeze(0).mul(1.0 / 127.0))
+            if bias is not None:
+                output = output.add_(bias)
+
+        # we apply the fused bias here
```
borzunov, line 254: The comment is out-of-place, please move it right after L237 as in the original code :)

borzunov (approved):

PR #114, opened 2022-12-02
Measure throughput for different configs, devices, and dtypes separately

borzunov:


src/petals/server/throughput.py
```diff
@@ -104,21 +103,43 @@ def measure_network_rps(config: BloomConfig) -> float:
...
-
+    logger.info(
+        f"Compute throughput ({_get_device_name(device)}, {_get_dtype_name(dtype, load_in_8bit)}): "
+        f"{device_rps:.2f} RPS"
+    )
     return device_rps
+
+
+def _get_device_name(device: torch.device) -> str:
+    return f"{torch.cuda.get_device_name(0)} GPU" if device == "cuda" else "CPU"
```
justheuristic, line 141:
```suggestion
    device = torch.device(device)  # compatibility with string device names
    return f"{torch.cuda.get_device_name(device.index)} GPU" if device.type == "cuda" else "CPU"
```

Nit
image
image

justheuristic (approved):

PR #115, opened 2022-12-02
Make server use smart defaults

borzunov:
Summary:

```python
parser.add_argument('--attn_cache_size', type=str, default=None,
                    help='The size of GPU memory allocated for storing past attention keys/values between inference steps. '
                         'Examples: 500MB, 1.2GB, 1073741824 (bytes). Note that 1KB != 1KiB here. '
                         'Default: 0.5GiB * num_blocks * hidden_size / 14336. '
                         'The latter is the hidden size of the bigscience/bloom-petals model.')

parser.add_argument('--request_timeout', type=float, required=False, default=3 * 60,
                    help='Timeout (in seconds) for the whole rpc_forward/rpc_backward/rpc_forward_stream/rpc_backward_stream request')
parser.add_argument('--session_timeout', type=float, required=False, default=30 * 60,
                    help='Timeout (in seconds) for the whole inference session')
parser.add_argument('--step_timeout', type=float, required=False, default=60,
                    help="Timeout (in seconds) for waiting the next step's inputs inside an inference session")

parser.add_argument('--load_in_8bit', type=bool, default=None,
                    help="Convert the loaded model into mixed-8bit quantized model. Default: True if GPU is available")
```

### DIFF
--- a/src/petals/server/throughput.py
+++ b/src/petals/server/throughput.py
@@ -6,6 +6,7 @@ import tempfile
 import time
 from hashlib import sha256
 from pathlib import Path
+from typing import Union
 
 import torch
 from hivemind.utils.logging import get_logger, use_hivemind_log_handler
@@ -26,7 +27,7 @@ DEFAULT_LOCK_PATH = Path(tempfile.gettempdir(), "petals", "throughput.lock")
 def get_host_throughput(
     config: BloomConfig,
     device: torch.device,
-    torch_dtype: torch.dtype,
+    dtype: Union[str, torch.dtype],
     *,
     load_in_8bit: bool,
     force_eval: bool = False,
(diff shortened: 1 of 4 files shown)


justheuristic (approved):

src/petals/server/throughput.py
```diff
@@ -55,7 +56,7 @@ def get_host_throughput(
             cache = {}
 
         if cache_key not in cache:
-            cache[cache_key] = measure_throughput_info(config, device, torch_dtype, load_in_8bit=load_in_8bit)
```
justheuristic, line 58: shoudln't torch_dtype be a part of config?

justheuristic (approved):

PR #119, opened 2022-12-02
Choose --num_blocks for bigscience/bloom-petals automatically

borzunov:


src/petals/server/server.py
```diff
@@ -167,6 +170,22 @@ def __init__(
 
         self.stop = threading.Event()
 
+    def _choose_num_blocks(self) -> int:
+        assert self.converted_model_name_or_path != "bigscience/bloom-petals", \
```
borzunov, line 174: We can make a generic solution for all models later (e.g., by creating a fake block inside `with accelerate.no_init_weights():` and counting its parameters)

src/petals/server/server.py
```diff
@@ -167,6 +170,25 @@ def __init__(
...
+            "If you use a model other than bigscience/bloom-petals, please specify --num blocks manually"
+        assert self.device.type == "cuda", "If you run a non-GPU server, please specify --num_blocks manually"
+
+        gib = 1024**3
+        total_memory_gib = torch.cuda.get_device_properties(self.device).total_memory / gib
+        block_size_gib = 176 / 70 + 0.5
```
justheuristic, line 181: [nit] model has 172.6 parameters in transformer blocks, the rest is in embeddings
[nit] a generic estimate is  hid_size * 2 * 12 * (1.1 if load_in_8bit else 2)

justheuristic (approved):

PR #126, opened 2022-12-03
Fix typos with codespell

mryab:


src/petals/bloom/ops.py
```diff
@@ -116,7 +116,7 @@ def dropout_add(x, residual, prob, training):
     Args:
         x (`torch.tensor`, *required*):
             input tensor
-        residual (`torch.tensor`, *rquired*):
+        residual (`torch.tensor`, *required*):
```
justheuristic, line 119: this might be on HF

justheuristic (approved):

borzunov (approved):

src/petals/bloom/ops.py
mryab: Yes, but it got fixed in https://github.com/huggingface/transformers/commit/b69a62d579ded31d490e0c9a01bf9ecb81cb9b65 (by the way, we might want to pull upstream changes at some point)

PR #127, opened 2022-12-03
Avoid synchronous updates, ban peers based on request outcome

justheuristic:


src/petals/client/routing/sequence_manager.py
```diff
@@ -158,6 +173,24 @@ def _update(self):
                 logger.log(traceback_level, "See detailed traceback below:", exc_info=True)
                 time.sleep(delay)
 
+    def on_request_failure(self, peer_id: PeerID):
+        """remove a given peer from the routing table. If the routing is no longer possible, trigger an update"""
+        logger.debug(f"Banning {peer_id}")
```
borzunov, line 178:
```suggestion
        logger.info(f"Peer {peer_id} does not respond, banning it for {self.ban_timeout} sec")
```

PR #132, opened 2022-12-03
Allow .generate() to reuse existing inference session

borzunov:
Allows to write chatbots like this:

```python
with model.inference_session(max_length=512) as sess:
    while True:
        prompt = f"Human: {input('Human: ')}\nAI:"
        prefix = tokenizer(prompt, return_tensors="pt")["input_ids"].cuda()
        print("AI:", end='', flush=True)
        while True:
            outputs = model.generate(prefix, max_new_tokens=1, do_sample=True, top_p=0.9, temperature=0.75, session=sess)
            outputs = tokenizer.decode(outputs[0, -1:])
            print(outputs, end='', flush=True)
            if "\n" in outputs:
                break
            prefix = None  # We pass the prefix only for the 1st token of the bot's response
```

How it works:

```
Human: Hi!
AI: Hello.
Human: How are you doing?
AI: Fine, thank you. And you?
Human: Fine too. Please tell me, what do you think about humans? 
AI: I have a lot of respect for humans. I have learned a lot from you.
```

### DIFF
--- a/src/petals/client/remote_generation.py
+++ b/src/petals/client/remote_generation.py
@@ -106,10 +106,13 @@ class RemoteGenerationMixin:
         if inputs is not None:
             assert isinstance(inputs, torch.Tensor) and inputs.ndim == 2, "inputs must be a 2d tensor [batch, length]"
             if session is not None and session.last_token_id is not None:
-                inputs = torch.cat([
-                    torch.tensor([[session.last_token_id]] * num_beams, dtype=torch.long, device=self.device),
-                    inputs,
-                ], dim=1)
+                inputs = torch.cat(
+                    [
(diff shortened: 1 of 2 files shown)


src/petals/client/remote_generation.py
```diff
@@ -167,7 +189,7 @@ def generate(
 
                 outputs.append(last_token_id)
                 seq_idx += 1
-                if torch.all(last_token_id == eos_token_id) or len(outputs) > max_new_tokens:
+                if torch.all(last_token_id == eos_token_id) or len(outputs) >= max_length:
```
borzunov, line 192: TODO: Revert this, it was actually correct

PR #141, opened 2022-12-08
Use common folder for all caches, make it a volume in Dockerfile

borzunov:


src/petals/utils/disk_cache.py
```diff
@@ -0,0 +1,5 @@
+import os
+from pathlib import Path
+
+
+DEFAULT_CACHE_DIR = os.getenv("PETALS_CACHE", Path(Path.home(), ".cache", "petals"))
```
mryab, line 4:
```suggestion
DEFAULT_CACHE_DIR = os.getenv("PETALS_CACHE", Path.home() / ".cache" / "petals")
```

src/petals/server/throughput.py
```diff
@@ -15,31 +14,32 @@
...
 def get_host_throughput(
     config: BloomConfig,
     device: torch.device,
     dtype: Union[str, torch.dtype],
     *,
     load_in_8bit: bool,
     force_eval: bool = False,
-    cache_path: str = DEFAULT_CACHE_PATH,
-    lock_path: str = DEFAULT_LOCK_PATH,
+    cache_dir: Optional[str] = None,
```
mryab, line 30:
```suggestion
    cache_dir: Optional[Union[str, Path]] = None,
```

mryab (approved):

src/petals/utils/disk_cache.py
borzunov: I don't like that this notation doesn't work for all strings and less readable for Python amateurs

src/petals/server/throughput.py
borzunov: `Path` has all interfaces of `str` (duck typing), so I'll leave my variant to not overcomplicate the code

PR #143, opened 2022-12-09
Add local tensor-parallel fwd/bwd

justheuristic:
This pull request adds an option to run Petals server on multiple local GPUs. It uses https://github.com/BlackSamorez/tensor_parallel

- [x] tensor_parallel works with forward/backward
- [x] tensor_parallel works with inference (use_cache=True)
- [x] exact match output, grad w.r.t. input, inference caches
- [x] works with odd numbers of GPUs
- [x] del block frees memory from all GPUs (for rebalancing)
- [x] stability tests
- [x] add to CI tests
- [x] determine num_blocks
- [x] measure throughput correctly

### Benchmark

https://gist.github.com/justheuristic/149ccfbf903a847cbaa09dbe59965bd9

### Overnight sanity checks:

- 8bit approximation error same as in main (mean~=2% q0.9~=5%)
    - TP=1, 2, 3 (see screenshots above)
- forward, grad w.r.t. input and inference exact match with main with TP=1
- `>=`80% GPU utilization with 3x 1080ti, batch = 8 tokens
- throughput measured with and without TP
- TP on 1080Tis has near-linear speedup comparable to the benchmarks (see first message)


tests/test_remote_sequential.py
```diff
@@ -59,7 +59,7 @@ def test_remote_sequential():
     assert not torch.allclose(test_inputs.grad, full_grad, rtol=0, atol=1e-2), "compression was not used"
     assert abs(approx_outputs - full_outputs).mean() < 0.01
     absmax = abs(full_grad).max()
-    assert abs(test_inputs.grad / absmax - full_grad / absmax).mean() < 0.01
+    assert abs(test_inputs.grad / absmax - full_grad / absmax).mean() < 0.05
```
justheuristic, line 62: Not related to TP: i caught a random error 2 times out of 90 test runs

tests/test_remote_sequential.py
```diff
@@ -118,7 +118,7 @@ def test_remote_sequential_prompts(batch_size=2, seq_len=5, pre_seq_len=3):
         block = load_pretrained_block(MODEL_NAME, block_index=block_index, torch_dtype=torch.float32)
         (outputs_ref,) = block(outputs_ref)
 
-    assert torch.allclose(outputs_ref, outputs)
+    assert torch.allclose(outputs_ref, outputs, atol=1e-5)
```
justheuristic, line 121: This one *is* related to tensor-parallelism

Previously, the float32 no-compression server would run exactly the same sequence of operations and was exactly equal.
Now, we compare non-tensor-parallel model to a tensor-parallel one. It is still allcose, but the summation order slightly affects the outputs.

.github/workflows/run-tests.yaml
```diff
@@ -87,15 +87,15 @@ jobs:
           sleep 10 # wait for initial servers to declare blocks, then let server decide which blocks to serve
 
           python -m petals.cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 0:6 \
-            --initial_peers $INITIAL_PEERS --throughput 1 --torch_dtype float32 &> server3.log &
+            --initial_peers $INITIAL_PEERS --throughput 1 --tensor_parallel_devices cpu cpu --torch_dtype float32 &> server3.log &
```
justheuristic, line 90: The intent is to make sure that there is at least one tensor-parallel server in every chain

src/petals/utils/convert_block.py
```diff
@@ -0,0 +1,97 @@
...
+from hivemind.utils.logging import get_logger, use_hivemind_log_handler
+
+from petals.bloom.block import WrappedBloomBlock
+from petals.utils.linear8bitlt_patch import CustomLinear8bitLt
+
+use_hivemind_log_handler("in_root_logger")
+logger = get_logger(__file__)
+
+
+def replace_8bit_linear(module: nn.Module, threshold=6.0) -> nn.Module:
```
justheuristic, line 19: moved from convert_8bit

src/petals/server/handler.py
```diff
@@ -352,8 +353,7 @@ async def _allocate_caches(
             total_size = 0
             backend = None
             for backend in backends:
-                num_heads = backend.module.self_attention.num_heads
-                head_dim = backend.module.self_attention.head_dim
+                num_heads, head_dim = backend.config.n_head, backend.config.hidden_size // backend.config.n_head
```
borzunov, line 356:
```suggestion
                num_heads = backend.config.n_head
                head_dim = backend.config.hidden_size // num_heads
```

src/petals/server/handler.py
```diff
@@ -20,6 +20,7 @@
 from hivemind.utils.asyncio import amap_in_executor, anext
 from hivemind.utils.logging import get_logger
 from hivemind.utils.streaming import split_for_streaming
+from transformers import BloomConfig
```
borzunov, line 23:
```suggestion
```

(unused import)

tests/test_tensor_parallel.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.mark.forked
+@pytest.mark.parametrize("custom_config", [True, False])
+@pytest.mark.parametrize("devices", [("cpu",) * 2, ("cpu",) * 3, ("cpu",) * 4])
+def test_tp_block(devices, custom_config):
+    block_index = random.randint(0, 10)
+    block = load_pretrained_block(MODEL_NAME, block_index=block_index, torch_dtype=torch.float32).to(devices[0])
+
+    tp_config = None
+    if custom_config:
+        tp_config = Config(
```
borzunov, line 20: Seems like this is the same config as in `convert_block.py`. Please make it a constant in `convert_block.py`, then import it here.

src/petals/utils/convert_block.py
```diff
@@ -0,0 +1,97 @@
...
+            ".*self_attention\.dense\.(weight|bias)": "split 0",
+            ".*mlp\.dense_h_to_4h\.(weight|bias)": "split 0",
+            ".*mlp\.dense_4h_to_h\.weight": "split 1",
+            ".*mlp\.dense_4h_to_h\.bias": "scale",
+        },
+        input_rules={},
+        output_rules={
+            ".*self_attention\.query_key_value": {0: "gather -1"},
+            ".*self_attention\.dense": {0: "gather -1"},
+            ".*mlp\.dense_4h_to_h$": {0: "sum"},
```
borzunov, line 68:
```suggestion
            r".*self_attention\.query_key_value": {0: "gather -1"},
            r".*self_attention\.dense": {0: "gather -1"},
            r".*mlp\.dense_4h_to_h$": {0: "sum"},
```

src/petals/utils/convert_block.py
```diff
@@ -0,0 +1,97 @@
...
+
+def make_tensor_parallel(block: WrappedBloomBlock, devices: Sequence[torch.device], output_device: torch.device):
+    assert isinstance(block, WrappedBloomBlock)
+    tp_config = tp.Config(
+        state_rules={
+            ".*self_attention\.query_key_value\.(weight|bias)": "split 0",
+            ".*self_attention\.dense\.(weight|bias)": "split 0",
+            ".*mlp\.dense_h_to_4h\.(weight|bias)": "split 0",
+            ".*mlp\.dense_4h_to_h\.weight": "split 1",
+            ".*mlp\.dense_4h_to_h\.bias": "scale",
```
borzunov, line 62:
```suggestion
            r".*self_attention\.query_key_value\.(weight|bias)": "split 0",
            r".*self_attention\.dense\.(weight|bias)": "split 0",
            r".*mlp\.dense_h_to_4h\.(weight|bias)": "split 0",
            r".*mlp\.dense_4h_to_h\.weight": "split 1",
            r".*mlp\.dense_4h_to_h\.bias": "scale",
```

You were lucky since `"\." == r"\."` but this is not true for all regexps.

tests/test_tensor_parallel.py
```diff
@@ -0,0 +1,59 @@
...
+                r".*self_attention\.dense\.(weight|bias)": "split 0",
+                r".*mlp\.dense_h_to_4h\.(weight|bias)": "split 0",
+                r".*mlp\.dense_4h_to_h\.weight": "split 1",
+                r".*mlp\.dense_4h_to_h\.bias": "scale",
+            },
+            input_rules={},
+            output_rules={
+                ".*self_attention\.query_key_value": {0: "gather -1"},
+                ".*self_attention\.dense": {0: "gather -1"},
+                ".*mlp\.dense_4h_to_h$": {0: "sum"},
```
borzunov, line 32:
```suggestion
                r".*self_attention\.query_key_value": {0: "gather -1"},
                r".*self_attention\.dense": {0: "gather -1"},
                r".*mlp\.dense_4h_to_h$": {0: "sum"},
```

src/petals/server/throughput.py
```diff
@@ -112,14 +112,18 @@ def measure_compute_rps(
     dtype: torch.dtype,
     *,
     load_in_8bit: bool,
+    tensor_parallel_devices: Sequence[torch.device],
```
borzunov, line 115: Please add `tensor_parallel_devices` param to `get_host_throughput()`, pass it there, and also add it to `cache_key` (if present)

borzunov (changes requested):
Overall it's wonderful, let's address some nits (most importantly, measure throughput for TP case correctly).

src/petals/utils/convert_block.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+            module._modules[n].weight = bnb.nn.Int8Params(
+                module.weight.data, requires_grad=False, has_fp16_weights=False
+            ).to(module.weight.dtype)
+            module._modules[n].bias = module.bias
+    return module
+
+
+def make_tensor_parallel(block: WrappedBloomBlock, devices: Sequence[torch.device], output_device: torch.device):
+    assert isinstance(block, WrappedBloomBlock)
```
borzunov, line 55:
```suggestion
    assert isinstance(block, (WrappedBloomBlock, CustomLinear8bitLt))
```

borzunov:
Some issues I met after trying to run that on real model:

**Issue 1**

```python
  File "/home/user/miniconda3/envs/jupyter/lib/python3.10/site-packages/tensor_parallel/slicer_wrapper.py", line 225, in process_state_
    logger.warning(f"The following patterns in state_rules were unused: {sorted(unused_patterns)}")
TypeError: '<' not supported between instances of 're.Pattern' and 're.Pattern'
```

**Issue 2**

Happens unless you specify `--device cuda:0 --tensor_parallel_devices cuda:0 cuda:1` (since the default device is just `cuda` and they are not compared properly):

```python
  File "/home/user/miniconda3/envs/jupyter/lib/python3.10/site-packages/tensor_parallel/tensor_parallel.py", line 47, in __init__                                                                           
    self.output_device_index = self.devices.index(output_device) if output_device is not None else 0                                                                                                        
ValueError: tuple.index(x): x not in tuple           
```

tests/test_tensor_parallel.py
justheuristic: moved to core module (tensor_parallel)

src/petals/server/throughput.py
justheuristic: done, thanks!

.github/workflows/run-tests.yaml
```diff
@@ -86,16 +86,16 @@ jobs:
 
           sleep 10 # wait for initial servers to declare blocks, then let server decide which blocks to serve
 
-          python -m petals.cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 0:6 \
+          python -m petals.cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 0:5 \
             --initial_peers $INITIAL_PEERS --throughput 1 --torch_dtype float32 &> server3.log &
           SERVER3_PID=$!
 
-          python -m petals.cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 4:16 \
+          python -m petals.cli.run_server --converted_model_name_or_path $MODEL_NAME --block_indices 4:14 \
```
justheuristic, line 93: slightly less blocks; why: we're using about 95% of total memory for test_full_model, this ensures that we don't go overboard

justheuristic:
@borzunov please make another pass when you're ready

tests/test_remote_sequential.py
```diff
@@ -39,10 +40,10 @@ def test_remote_sequential():
     assert hidden.shape == test_inputs.shape
     assert hidden.requires_grad
     second_half_outputs = second_half(hidden)
-    assert torch.allclose(second_half_outputs, full_outputs)
+    assert torch.allclose(second_half_outputs, full_outputs, atol=1e-4)
```
justheuristic, line 43: related to TP: TP partial matrix products are added up in non-deterministic order. The 1e-4 tolerance is chosen because (1) errors accumulate for 24 layers and (2) we need some leeway to make sure tests aint flappy.

tests/test_tensor_parallel.py
borzunov: Can we import it here then?

src/petals/server/throughput.py
borzunov: You haven't added it to `cache_key`. Now, throughput won't be measured again if one enables tensor parallelism, the server will use the old throughput. Please fix this.

src/petals/cli/run_server.py
```diff
@@ -131,6 +131,7 @@ def main():
     parser.add_argument('--load_in_8bit', type=str, default=None,
                         help="Convert the loaded model into mixed-8bit quantized model. "
                              "Default: True if GPU is available. Use `--load_in_8bit False` to disable this")
+    parser.add_argument("--tensor_parallel_devices", nargs='+', default=[])  # disabled by default
```
borzunov, line 134: Seems like this feature is enabled by default, so please remove the comment. Also, please add the help string.

src/petals/server/throughput.py
```diff
@@ -129,14 +135,17 @@ def measure_compute_rps(
...
     n_tokens: int = 16,
     n_steps: int = 500,
 ) -> float:
+    if not tensor_parallel_devices:
+        tensor_parallel_devices = (device,)
```
borzunov, line 143: The server does not enable tensor-parallelism if we have only one device. Please behave here similarly.

E.g., if tensor-parallelism breaks but the server works otherwise, this code will fail (even though the server would have worked).

src/petals/server/task_pool.py
```diff
@@ -101,7 +101,7 @@ def shutdown(self, timeout: float = 3):
             logger.warning(f"{self.__class__.__name__} failed to shut down gracefully, sending SIGTERM")
             self.terminate()
 
-    def submit_task(self, *args: torch.Tensor, priority: float = 0.0) -> MPFuture:
+    def submit_task(self, *args: Union[torch.Tensor, Any], priority: float = 0.0) -> MPFuture:
```
borzunov, line 104:
```suggestion
    def submit_task(self, *args: Any, priority: float = 0.0) -> MPFuture:
```

src/petals/cli/run_server.py
borzunov, line 134:
```suggestion
    parser.add_argument("--tensor_parallel_devices", nargs='+', default=None)
```

Please add a help string here. If you want tensor-parallelism to be enabled by default (I think I'd prefer it), change the default to `None` and remove the comment.

src/petals/server/throughput.py
justheuristic: Discussion log:
- i complained that not wrapping with TensorParallel will complicate the code in backend.py, handler.py as follows:
```
If isinstance(backend.module, WrappedBloomBlock):
   do_thing()
elif isinstance(backend.module, TensorParallel):
   do_another_thing()
else:
   raise RuntimeError(...)
```

- i checked that TensorParallel with len(devices) <= 1 calls the model normally, without cross-device stuff
- @borzunov suggested that if we do wrap with TensorParallel, we should wrap the same way everywhere (e.g. server.py, throughput.py)

justheuristic:
### Sanity check: layer error

- __reference__ load module on CPU, run in bfloat16 (slowly)
- __baseline__ single GPU, quantize to 8-bit (old code)

image

- __TP=2__ two 1080ti GPUs, each is quantized as in `server.py:421`
image

- __TP=3__ three 1080ti GPUs, otherwise the same

image

src/petals/server/backend.py
```diff
@@ -48,44 +64,60 @@ def __init__(self, *args, memory_cache: MemoryCache, backend_dtype: torch.dtype,
             self.kwargs_schema,
         )
 
+    def get_inference_cache_descriptors(self, batch_size: int, max_length: int) -> Tuple[TensorDescriptor, ...]:
```
borzunov, line 67:
```suggestion
    def get_inference_cache_descriptors(self, batch_size: int, max_length: int) -> List[Tuple[TensorDescriptor, TensorDescriptor]]:
```

src/petals/server/backend.py
```diff
@@ -48,44 +64,60 @@ def __init__(self, *args, memory_cache: MemoryCache, backend_dtype: torch.dtype,
...
+        """Create tensor descriptors for attention cache tensors used during inference_step"""
+        head_dim = self.config.hidden_size // self.config.n_head
+        cache_tensors = []
+        for device, num_heads in zip(self.module.devices, self.shard_num_heads):
+            keys = TensorDescriptor((batch_size, num_heads, head_dim, max_length), dtype=self.dtype, device=device)
+            values = TensorDescriptor((batch_size, num_heads, max_length, head_dim), dtype=self.dtype, device=device)
+            cache_tensors.extend((keys, values))
+        return tuple(cache_tensors)
```
borzunov, line 75:
```suggestion
        return cache_tensors
```

src/petals/server/backend.py
borzunov:
```suggestion
    def get_inference_cache_descriptors(self, batch_size: int, max_length: int) -> List[Tuple[TensorDescriptor, TensorDescriptor]]:
```

justheuristic: This can be more than 2 tuples

src/petals/server/backend.py
```diff
@@ -48,44 +64,60 @@ def __init__(self, *args, memory_cache: MemoryCache, backend_dtype: torch.dtype,
             self.kwargs_schema,
         )
 
+    def get_inference_cache_descriptors(self, batch_size: int, max_length: int) -> Sequence[TensorDescriptor]:
```
borzunov, line 67:
```suggestion
    def get_inference_cache_descriptors(self, batch_size: int, max_length: int) -> Sequence[Tuple[TensorDescriptor, TensorDescriptor]]:
```

src/petals/cli/run_server.py
```diff
@@ -129,8 +129,14 @@ def main():
...
     parser.add_argument('--load_in_8bit', type=str, default=None,
-                        help="Convert the loaded model into mixed-8bit quantized model. "
+                        help="Convert the loaded transformer blocks into mixed-8bit quantized model. "
                              "Default: True if GPU is available. Use `--load_in_8bit False` to disable this")
+    parser.add_argument("--tensor_parallel_devices", nargs='+', default=None,
+                        help=
+                        "Split each block between the specified GPUs such that each device holds a portion of every "
+                        "weight matrix. The general idea is described in Section 3.2 of the AlexNet paper "
+                        "https://proceedings.neurips.cc/paper/2012/file/c399862d3b9d6b76c8436e924a68c45b-Paper.pdf "
+                        ", while the strategy for Transformers is described in https://arxiv.org/pdf/2104.04473.pdf .")
```
borzunov, line 139:
```suggestion
                        "Split each block between the specified GPUs such that each device holds a portion of every "
                        "weight matrix. See https://huggingface.co/transformers/v4.9.0/parallelism.html#tensor-parallelism")
```

I think this explanation is easier to comprehend for people with less research/academic background

src/petals/server/server.py
```diff
@@ -469,7 +495,8 @@ def __init__(
             )
             for _ in range(num_handlers)
         ]
-        self.runtime = Runtime(self.module_backends, **kwargs)
+        self.runtime = Runtime(self.module_backends, device=None, **kwargs)
+        # note: we set device=None in runtime to avoid casting all modules to device 0 in .run; we'll cast them manually
```
borzunov, line 499:
```suggestion
        # note: We set device=None in runtime to avoid casting all modules to device 0 in runtime.run(). tensor_parallel will cast it for us as needed.
```

src/petals/server/throughput.py
```diff
@@ -149,8 +160,14 @@ def measure_compute_rps(
                 elapsed += time.perf_counter() - start_time
         device_rps = n_steps * n_tokens / elapsed
 
+
+    devices_repr = get_device_name(device)
+    if len(tensor_parallel_devices) > 1:
+        device_names = tuple(map(get_device_name, tensor_parallel_devices))
+        devices_repr = ', '.join([f'{count}x {name}' for name, count in Counter(device_names).most_common()])
```
borzunov, line 166:
```suggestion
        devices_repr = ', '.join(f'{count}x {name}' for name, count in Counter(device_names).most_common())
```

borzunov (approved):

PR #148, opened 2022-12-11
Update advanced notebooks

artek0chumak:
Update examples with:
- new interactive chat mode, where session is saved while the dialog
- add adapters in sst-2 example


examples/prompt-tuning-personachat.ipynb
```diff
@@ -274,30 +274,32 @@
...
-    "    bloom_answer = tokenizer.batch_decode(outputs)[0]\n",
-    "    bloom_answer = bloom_answer[len(dialog):].split(\"\\n\")[0]\n",
-    "    print(bloom_answer)\n",
-    "    dialog += f\"{bloom_answer}\\n-----\\n\""
+    "with model.inference_session(max_length=512) as sess:\n",
+    "    while True:\n",
+    "        user_phrase = input()\n",
+    "        if len(user_phrase) == 0:\n",
+    "            break\n",
+    "        dialog += f\"{user_phrase}\\n-----\\n\"\n",
```
borzunov, line 282: When using `.generate()` reusing the inference session, you should not pass the entire prefix to `.generate()` every time. You should only pass the new user's phrase (the previous phrase and the model's output are already remembered in the attention caches).

This way, you don't need to store the entire history in the `dialog` variable. You can only store the last input/output in this example.

Also, I encourage you to show the tokens in real time, since waiting for ~10 sec for a new response is quite unpleasant. Please note how it's done in [my example](https://colab.research.google.com/drive/1Ervk6HPNS6AYVr3xVdQnY5a-TjjmLCdQ?usp=sharing).

borzunov (changes requested):
Thanks for the PR! I have an important comment regarding using the new `.generate()` interface.

examples/prompt-tuning-personachat.ipynb
artek0chumak: Thank you for your comment! I will change the behaviour of an interactive dialogue, and make it like in your example.

borzunov (approved):

PR #152, opened 2022-12-13
Clean up disk space

borzunov:
**What's good in this PR?**

- The server (with rebalancing enabled) now works with any amount of free disk space, starting from 5 GiB.
- In case of 350+ GiB, it cashes the full model.
- In case of lesser amount of free space, it removes least recently used blocks once no free space is left.
- This works even in case of 5 GiB (the server loads a block, then removes it to free space for the next one).
- Optionally, a user may specify `--max_disk_space` to limit the space Petals use.
- This PR correctly handles multi-GPU machines with multiple Petals servers running by using a system-wide read-write lock. A block is never removed if it is being read by another server on the same machine.

**What can be improved in later PRs?**

1. On multi-GPU machines, Petals servers can simultaneously read from cache, but only one server can download a block at a time.
    - This limitation is introduced to keep the code simple. Otherwise, we'd need complex multi-process sync primitives, which will significantly increase the size of code base.
    - Still, I hope this limitation is reasonable. If the download speed is disk-bounded, we'd better download blocks sequentially (to reduce the number of the disk head's moves) than in parallel.
    - Otherwise, we could download blocks in parallel inside one server, which can be done in a separate PR and wouldn't require the complex sync primitives.
    - Finally, the recommended way of utilizing multi-GPU machines may be using one Petals server with tensor parallelism, so this feature becomes just unnecessary in this case.
  
2. Instead of loading blocks from `start` to `end`, we could load the cached blocks at first, then download new blocks (which may trigger removal of the old ones).
    - This saves traffic on machines with free disk space < 2 * gpu_memory. However, such setups are rare.

3. Currently, the LRU cache is built on top of block's access time. The downside is that if someone runs `grep -R ~`, the relevant access time of the files will be erased.
    - We could use a separate, explicit LRU cache, but this would significantly complicate the code (we need to serialize it, recover somehow if the file is corrupted, etc.). I think the cost of losing access time info (and using some extra traffic) in these rare cases is not large enough to justify the complications.


justheuristic (approved):
Removed some unused imports in src/petals/server/server. Otherwise, everything seems to be in order. Good job!

PR #156, opened 2022-12-15
Fix logging: do not duplicate lines, enable colors in Colab

borzunov:
Before this PR, logs contained duplicate lines in Colab, and the colors were disabled. This PR fixes it and does some refactoring of logging along the way. Now it's beautiful:

Screenshot 2022-12-15 at 09 11 23

### DIFF
--- a/src/petals/client/remote_model.py
+++ b/src/petals/client/remote_model.py
@@ -5,7 +5,7 @@ from typing import List, Optional
 import hivemind
 import torch
 import torch.nn as nn
-from hivemind.utils.logging import get_logger, loglevel, use_hivemind_log_handler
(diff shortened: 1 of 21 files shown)


src/petals/client/remote_model.py
```diff
@@ -21,13 +21,8 @@
 from petals.constants import PUBLIC_INITIAL_PEERS
 from petals.utils.misc import DUMMY
 
-use_hivemind_log_handler("in_root_logger")
 logger = get_logger(__file__)
 
-# We suppress asyncio error logs by default since they are mostly not relevant for the end user
```
borzunov, line 27: Moved to `petals/utils/logging.py`

PR #157, opened 2022-12-15
Switch to speedtest-cli

justheuristic:
This pullrequest removes custom speed_test code in favour of speedtest-cli module.
This is necessary to ensure that random warnings / print-outs do not mess with our outputs:
image

speedtest-cli is a fairly popular module trusted by many devs
image


src/petals/server/throughput.py
```diff
@@ -88,10 +88,30 @@ def measure_throughput_info(
 
 
 def measure_network_rps(config: BloomConfig) -> float:
-    proc = subprocess.run("python3 -m petals.cli.speed_test --json", shell=True, capture_output=True)
-    if proc.returncode != 0:
-        raise RuntimeError(f"Failed to measure network throughput (stdout: {proc.stdout}, stderr: {proc.stderr})")
-    network_info = json.loads(proc.stdout)
+    try:
+        import speedtest
```
mryab, line 92: The import should be on top of the module, if possible. This is not an optional dependency, and thus the module needs to fail as early as possible

src/petals/server/throughput.py
```diff
@@ -88,10 +88,30 @@ def measure_throughput_info(
...
+
+        s = speedtest.Speedtest()
+    except ImportError as e:
+        logger.error("Please `pip install speedtest-cli==2.1.3` or set throughput manually")
+        logger.exception(e)
+        raise e
```
mryab, line 98: If you raise the exception, there is no need to call logger.exception, as the traceback will be printed by the raised exception

src/petals/server/throughput.py
```diff
@@ -88,10 +88,30 @@ def measure_throughput_info(
...
+        )
+    try:
+        s.get_servers()
+        s.get_best_server()
+        s.download()
+        s.upload()
+        network_info = s.results.dict()
+    except Exception as e:
+        logger.error("Failed to measure network throughput:")
+        raise e
```
mryab, line 114:
```suggestion
    except:
        logger.error("Failed to measure network throughput:")
        raise
```

src/petals/server/throughput.py
justheuristic: thanks, fixed it

src/petals/server/throughput.py
justheuristic: fixed

src/petals/server/throughput.py
```diff
@@ -18,6 +17,20 @@
 
 logger = get_logger(__file__)
 
+try:
+    import speedtest
+except ImportError:
+    logger.error("Please `pip install speedtest-cli==2.1.3`")
+    raise
```
mryab, line 24:
```suggestion
    raise ImportError("Please `pip install speedtest-cli==2.1.3`")
```

mryab (approved):

src/petals/server/throughput.py
justheuristic: Respectfully disagree: it is possible that speedtest-cli throws ImportError because some other module failed to import inside of it. In this case, it is best to let user know the specific error (via `raise`)

justheuristic: nope, you were right
image

PR #160, opened 2022-12-16
Alloc inference cache as one contiguous buffer

borzunov:
If a client calls `rpc_inference()` for `N` blocks, server now allocates a single tensor for all caches instead of `N` tensors for each block's cache. This is useful because:

- **Bug:** If one block's cache size is less than `attn_cache_size`, but the total cache size is greater than `attn_cache_size`, the server will block until the `alloc_timeout` (60 sec) expires. Nobody else will not be able to call `rpc_inference()` during this time.
- **Optimization:** Allocating one large tensor decrease memory fragmentation.

The PR also fixes another **bug**: the max task size was not enforced to `rpc_inference()`, so a user could send very large requests and overload the server (see the comment).

### DIFF
--- a/src/petals/server/backend.py
+++ b/src/petals/server/backend.py
@@ -48,25 +48,25 @@ class TransformerBackend(ModuleBackend):
             self.kwargs_schema,
         )
 
-    def inference_step(self, cache_metadata: torch.IntTensor, *inputs: torch.Tensor) -> Tuple[torch.Tensor, ...]:
+    def inference_step(
+        self, hidden_states: torch.Tensor, hypo_ids: torch.LongTensor, cache_metadata: torch.LongTensor
+    ) -> Tuple[torch.Tensor, ...]:
         num_heads, head_dim = self.module.self_attention.num_heads, self.module.self_attention.head_dim
         with torch.inference_mode():
-            attention_cache_handle = int(cache_metadata[0, 0].item())
-            prefix_length = int(cache_metadata[0, 1].item())
-            (hidden_states, hypo_ids) = inputs
             assert (
                 hidden_states.ndim == 3
             ), "expected hidden states to be 3-dimensional: [batch_size, seq_len, hid_size]"
+            cache_handle, rel_index, prefix_length = map(int, cache_metadata[0])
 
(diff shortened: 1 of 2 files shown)


src/petals/server/backend.py
```diff
@@ -48,25 +48,28 @@ def __init__(self, *args, memory_cache: MemoryCache, backend_dtype: torch.dtype,
             self.kwargs_schema,
         )
 
-    def inference_step(self, cache_metadata: torch.IntTensor, *inputs: torch.Tensor) -> Tuple[torch.Tensor, ...]:
```
borzunov, line 51: Fixed bug: `PrioritizedTaskPool` calculates the "task size" using the first 2 dims of the 1st arg to this function. The 2 dims of cache_metadata were `(2, batch_size)` and didn't include `prefix_length`, so effectively the task size was not enforced for inference (a user could send an inference with prefix_length = 2048 and batch_size = 2048!).

src/petals/server/backend.py
```diff
@@ -48,25 +48,25 @@ def __init__(self, *args, memory_cache: MemoryCache, backend_dtype: torch.dtype,
...
             ), "expected hidden states to be 3-dimensional: [batch_size, seq_len, hid_size]"
+            cache_handle, rel_index, prefix_length = map(int, cache_metadata[0])
 
-            with self.memory_cache.use_cache(attention_cache_handle) as cache:
-                batch_size = cache.shape[1]
-                max_length = cache.numel() // (2 * batch_size * head_dim * num_heads)
-                assert isinstance(self.module, WrappedBloomBlock) and cache.shape[0] == 2 and cache.ndim == 3
+            with self.memory_cache.use_cache(cache_handle) as cache:
+                batch_size = cache.shape[2]
+                max_length = cache.shape[-1] // (head_dim * num_heads)
```
borzunov, line 63: `cache` shape has changed from `(2, batch_size, ...)` to `(n_blocks_in_request, 2, batch_size, ...)`

justheuristic (approved):
LGTM. I'm a bit concerned that allocation gets more complicated -- especially after we make this parallel. That said, it does remove cases two clients lock each other out of cache.

PR #166, opened 2022-12-28
Correct grammar in readme

vadi2:


borzunov (approved):

justheuristic:
The reason tests fail is that they don't have access to bloom_testing secret. I'm merging this PR regardless.

PR #179, opened 2023-01-04
Disable chunked_forward() on AVX512 CPUs

borzunov:
See benchmarks in #176.

### DIFF
--- a/src/petals/bloom/modeling_utils.py
+++ b/src/petals/bloom/modeling_utils.py
@@ -4,9 +4,11 @@ Based on https://github.com/huggingface/transformers/commit/ca2a55e9dfb245527b5e
 See commit history for authorship.
 """
 
+import psutil
 import torch
 import torch.nn.functional as F
 import torch.utils.checkpoint
+from cpufeature import CPUFeature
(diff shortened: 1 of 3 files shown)


src/petals/bloom/modeling_utils.py
```diff
@@ -59,17 +68,25 @@ def forward(self, hidden_states):
...
+        chunked_forward_step: provides trade-off between efficiency and extra memory consumption.
         """
-        assert self.chunk_size > 0, "Chunk size for chunked forward must be positive"
+        assert self.chunked_forward_step > 0, "Chunk size for chunked forward must be positive"
+
+        if not self._bf16_warning_shown:
+            if self.word_embeddings.weight.numel() * 4 < 0.9 * psutil.virtual_memory().total:
+                logger.warning(
+                    "Running the client with dtype bfloat16 on CPU may be slow, since your CPU doesn't support AVX512. "
+                    "Consider loading the model with torch_dtype='float32'"
```
borzunov, line 79: In future, we can also suggest to use `fast_but_approximate=True` here, which will run faiss under the hood for greedy and top_k generation

PR #186, opened 2023-01-08
Support libp2p relays for NAT traversal

Vahe1994:
- Added relay options to servers
- Enabled relay options by default
- Changed hivemind version to 1.1.5
- Moved reachability check to be performed after blocks are loaded

### DIFF
--- a/src/petals/cli/run_server.py
+++ b/src/petals/cli/run_server.py
@@ -12,6 +12,9 @@ from petals.server.server import Server
 logger = get_logger(__file__)
 
 
+TRUE_CONSTANTS = ["TRUE", "1"]
+
+
 def main():
     # fmt:off
     parser = configargparse.ArgParser(default_config_files=["config.yml"],
@@ -127,8 +130,9 @@ def main():
     parser.add_argument("--mean_balance_check_period", type=float, default=60,
                         help="Check the swarm's balance every N seconds (and rebalance it if necessary)")
 
-    parser.add_argument("--auto_relay", action='store_true', help="Enabling relay for NAT traversal")
-    parser.add_argument('--no-auto_relay', dest='auto_relay', action='store_false')
+    parser.add_argument("--use_auto_relay", type=str, default="True",
+                        help="Look for libp2p relays for NAT traversal. "
+                             "Use `--use_auto_relay False/True` to disable/enable this")
     parser.add_argument("--use_auth_token", action='store_true', help="auth token for from_pretrained")
 
     parser.add_argument('--load_in_8bit', type=str, default=None,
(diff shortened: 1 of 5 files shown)


src/petals/cli/run_server.py
```diff
@@ -127,7 +127,10 @@ def main():
     parser.add_argument("--mean_balance_check_period", type=float, default=60,
                         help="Check the swarm's balance every N seconds (and rebalance it if necessary)")
 
+    parser.add_argument("--auto_relay", action='store_true', help="Enabling relay for NAT traversal")
```
mryab, line 130:
```suggestion
    parser.add_argument("--auto_relay", action='store_true', help="Enable relay for NAT traversal")
```

src/petals/server/server.py
```diff
@@ -78,6 +78,8 @@ def __init__(
         load_in_8bit: Optional[bool] = None,
         tensor_parallel_devices: Optional[Sequence[torch.device]] = None,
         skip_reachability_check: bool = False,
+        use_relay: bool = True,
```
mryab, line 81: It's best to either remove default argument values of to remove these arguments completely: we might forget to change defaults here in the future, and required values will be passed to kwargs anyway

mryab (approved):

src/petals/server/server.py
borzunov: I see your point but disagree, since the convention in Petals is that all `Server` defaults match to the defaults of `run_server.py`.

Vahe1994: I will agree with @borzunov here. Here is my arguments:
1. use_relay will not be passed from run_server and we want it by default to be True
2. it is nice to see in the arguments all parameters that matters
3.  usually , it is not a good idea to be dependent on default argument from another library . They could be changed without notice and can lead to strange behavior

mryab: I see your point about explicitly indicating arguments for creation of Server, though it somewhat contradicts the existence of **kwargs in init. My primary concern is that we should strive to have consistent defaults across different locations: one way to do this in an error-proof way would be to declare a common constant with the default value and use it in both locations. Besides, petals-cli is a part of Petals, so these files belong to the same library

borzunov:
Thanks for the PR!

I have questions about relays, need to talk to @justheuristic before I approve this code.

src/petals/server/server.py
borzunov: We would need to create constants for all defaults in this case (tens of them). I think this is a more general problem that should be addressed outside of this PR (maybe we should use smth like reflection).

borzunov (changes requested):
Thanks for the PR!

I have questions about relays, I need to talk to @justheuristic before we ship this code.

src/petals/server/reachability.py
```diff
@@ -0,0 +1,38 @@
+import time
+
+import requests
+from hivemind.utils.logging import get_logger
+
+logger = get_logger(__file__)
+
+
+def check_reachability(peer_id, wait_time: float = 7 * 60, retry_delay: float = 15) -> None:
```
borzunov, line 9: Moved from `server.py`

PR #189, opened 2023-01-10
Bump version to 1.1.0

borzunov:


borzunov:
Closed since tests fail on branches with dots in their name.

PR #191, opened 2023-01-10
Return available cache size in rpc_info()

justheuristic:
This PR makes servers return their free cache (in tokens * layers to make it compression-agnostic)

To be used when calling make_sequence(optimize="inference")


src/petals/server/handler.py
```diff
@@ -378,6 +380,23 @@ def _log_request(
...
             logger.warning(f"{message}: {warning}")
 
+    async def rpc_info(self, request: runtime_pb2.ExpertUID, context: P2PContext) -> runtime_pb2.ExpertInfo:
+        """Return metadata about stored block uids and current load"""
+        rpc_info = {}
+        if request.uid:
+            backend = self.module_backends[request.uid]
+            rpc_info.update(self.module_backends[request.uid].get_info())
+        else:
+            backend = next(iter(self.module_backends.values()))
```
justheuristic, line 390:
```suggestion
            backend = next(iter(self.module_backends.values()))
            # not saving keys to rpc_info since user did not request any uid
```

src/petals/server/handler.py
```diff
@@ -378,6 +380,23 @@ def _log_request(
...
+
+        cache_bytes_left = max(0, backend.memory_cache.max_size_bytes - backend.memory_cache.current_size_bytes)
+        bits_per_token = 2 * backend.args_schema[0].shape[-1] * torch.finfo(backend.args_schema[0].dtype).bits // 8
+        # TODO does not account for tensor parallelism! switch to backend.get_descriptors once TP is merged, save as property to avoid recomputation
+        if CACHE_TOKENS_AVAILABLE in rpc_info:
+            logger.error(f"Block rpc_info dict has a reserved field {CACHE_TOKENS_AVAILABLE} : {rpc_info}")
```
justheuristic, line 396:
```suggestion
            raise RuntimeError(f"Block rpc_info dict has a reserved field {CACHE_TOKENS_AVAILABLE} : {rpc_info}")
```

borzunov (approved):

PR #198, opened 2023-01-11
Ignore network RPS if we failed to measure it

borzunov:


src/petals/server/throughput.py
```diff
@@ -101,25 +101,25 @@ def measure_throughput_info(
...
+    except Exception:
+        logger.warning("Failed to measure network throughput:", exc_info=True)
+        logger.warning("Proceeding with the compute throughput only")
+    return result
+
+
+def measure_network_rps(config: BloomConfig) -> Optional[float]:
+    s = speedtest.Speedtest()
+    s.get_servers()
+    s.get_best_server()
```
Vahe1994, line 119: here algorithm can be stuck for a long time

Vahe1994 (approved):

src/petals/server/throughput.py
borzunov: You're right, this PR won't fix this. I'd suggest to still merge it, since it'll fix test flaps like this:

PR #206, opened 2023-01-12
Provide attention mask into the transformer layers

artek0chumak:
Replace assert about not working attention_mask logic with actual attention mask transfer.

### DIFF
--- a/src/petals/client/inference_session.py
+++ b/src/petals/client/inference_session.py
@@ -78,6 +78,7 @@ class _ServerInferenceSession:
     def step(
         self,
         new_hidden_states: torch.Tensor,
+        attention_mask: torch.Tensor,
         prompts: Optional[torch.Tensor] = None,
         hypo_ids: Optional[torch.Tensor] = None,
     ) -> torch.Tensor:
@@ -104,7 +105,7 @@ class _ServerInferenceSession:
             assert hypo_ids.dtype == torch.int64
 
         # serialize inputs and put them into the queue
-        inputs = (new_hidden_states, prompts, hypo_ids)
+        inputs = (new_hidden_states, attention_mask, prompts, hypo_ids)
         outputs_serialized = RemoteExpertWorker.run_coroutine(
             self._step(
                 runtime_pb2.ExpertRequest(
@@ -212,7 +213,9 @@ class InferenceSession:
         assert not self._closed and not self._chosen_spans
         return self
 
-    def step(self, inputs: torch.Tensor, prompts: Optional[torch.Tensor] = None, **kwargs) -> torch.Tensor:
+    def step(
+        self, inputs: torch.Tensor, attention_mask: torch.Tensor, prompts: Optional[torch.Tensor] = None, **kwargs
+    ) -> torch.Tensor:
         assert not self._closed
         if torch.is_grad_enabled():
             logger.warning("Running inference session with grad enabled. Gradients will *not* be propagated correctly.")
@@ -226,6 +229,7 @@ class InferenceSession:
         inputs_device = inputs.device
         inputs_dtype = inputs.dtype
         inputs = inputs.cpu()
+        attention_mask = attention_mask.cpu()
         prompts = prompts.cpu()
 
         n_input_tokens = inputs.shape[1]
@@ -294,7 +298,7 @@ class InferenceSession:
                     else:
                         inputs = inputs[:, -n_input_tokens:]  # No need to pass prefix further
 
-                    outputs = session.step(inputs, prompts[span.start : span.end], **kwargs)
+                    outputs = session.step(inputs, attention_mask, prompts[span.start : span.end], **kwargs)
                     assert (
                         inputs.shape == outputs.shape
                     ), f"Shape mismatch: inputs.shape={inputs.shape}, outputs.shape={outputs.shape})"
--- a/src/petals/client/inference_session.py
+++ b/src/petals/client/inference_session.py
@@ -107,9 +107,6 @@ class _ServerInferenceSession:
         if attention_mask is None:
             attention_mask = DUMMY
 
-        if attention_mask is None:
-            attention_mask = DUMMY
-
         # serialize inputs and put them into the queue
         inputs = (new_hidden_states, attention_mask, prompts, hypo_ids)
         outputs_serialized = RemoteExpertWorker.run_coroutine(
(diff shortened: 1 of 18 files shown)


src/petals/client/remote_model.py
```diff
@@ -186,18 +184,24 @@ def forward(
...
+        batch_size = inputs_embeds.shape[0]
         if self.config.tuning_mode and "ptune" in self.config.tuning_mode:
-            batch_size = inputs_embeds.shape[0]
             prompts, intermediate_prompts = self.get_prompt(batch_size)
             inputs_embeds = torch.cat([prompts, inputs_embeds], dim=1)
 
         hidden_states = self.word_embeddings_layernorm(inputs_embeds)
         output_shape = input_shape + (hidden_states.size(-1),)
 
+        if attention_mask is None:
```
artek0chumak, line 195: Dub code

src/petals/client/inference_session.py
```diff
@@ -223,9 +236,16 @@ def step(self, inputs: torch.Tensor, prompts: Optional[torch.Tensor] = None, **k
         else:
             assert prompts.ndim == 4 and prompts.shape[0] == n_blocks
 
+        if attention_mask is None:
```
artek0chumak, line 239: Dub code

src/petals/client/routing/sequence_manager.py
```diff
@@ -333,7 +333,7 @@ def __del__(self):
 
 def maybe_log_traceback(exc: Exception):
     traceback_level = logging.DEBUG if str(exc) or isinstance(exc, asyncio.TimeoutError) else logging.WARNING
-    logger.log(traceback_level, "See detailed traceback below:", exc_info=True)
```
artek0chumak, line 336: Remove debug

src/petals/server/handler.py
```diff
@@ -156,6 +158,16 @@ async def rpc_inference(
                                 f" exceeds pre-allocated maximum {max_length}"
                             )
 
+                        if is_dummy(attention_mask):
```
artek0chumak, line 161: Dub code

src/petals/server/handler.py
```diff
@@ -323,9 +331,9 @@ def _serialize_grads(
     ) -> Sequence[runtime_pb2.Tensor]:
         """Serialize backward gradients w.r.t. inputs using either default schema or custom user-specified schema"""
         # Modify grad_inputs_schema to support grad_prompts
-        assert len(requested_backends[0].args_schema) == 1 and len(grads) in (1, 2)  # TODO generalize
+        assert len(requested_backends[0].args_schema) == 2 and len(grads) in (1, 2)  # TODO generalize
```
justheuristic, line 334: From conversation with @borzunov vs @artek0chumak vs @justheuristic 

We need a way to support (temporarily) a case where old client sends a request to the new server and vice versa
One way we could do that is to add code to `rpc_*` methods that checks for the right number of inputs and, if missing, adds an empty attention mask

src/petals/client/remote_forward_backward.py
```diff
@@ -88,8 +88,8 @@ async def run_remote_forward(
     # Modify forward_schema to support prompts
     args_schema, kwargs_schema = rpc_info["forward_schema"]
     # TODO: rm this assert when support arbitrary number of input tensors
-    assert len(args_schema) == 1 and len(inputs) == 2
-    forward_schema_with_prompts = (tuple(args_schema * len(inputs)), kwargs_schema)
+    assert len(args_schema) == 2 and len(inputs) == 3
```
justheuristic, line 91: from conversation of @artek0chumak vs @borzunov vs @justheuristic 

We need to support - temporarily - an edge case where old version client interacts with new version server and vice versa.
Since the code is temporary, it doesn't have to be perfectly elegant.

If you want, I can put together a PR into your PR to covert these two cases.

__Case 1:__ old client vs new server
- we can check for the number of inputs in each `rpc_*` method, and, if not given, create a dummy tensor right there and then
- in two weeks' time, we can remove this check

__Case 2:__ new client vs old server
- server expects exactly two arguments and does not have a futureproofing mechanism
- possible solution: simply do *not* send attention mask unless it is explicitly provided
   - in 2 weeks time, switch to sending mask all the time
- in theory, we could change the protocol so that client would check the server version before request. I would argue that it is an overkill. Then again, it's @artek0chumak 's choice.

src/petals/client/inference_session.py
```diff
@@ -103,8 +104,11 @@ def step(
             assert len(hypo_ids) == len(new_hidden_states)
             assert hypo_ids.dtype == torch.int64
 
+        if attention_mask is None:
+            attention_mask = DUMMY
+
```
justheuristic, line 109: If attention mask is None, inputs = (new_hidden_states, prompts, hypo_ids)
# TODO add a dummy attention mask here in 1.2.0

src/petals/client/remote_forward_backward.py
```diff
@@ -88,8 +88,8 @@ async def run_remote_forward(
...
+    forward_schema_with_prompts = ((args_schema[0], args_schema[1], args_schema[0]), kwargs_schema)
```
justheuristic, line 92: please be advised that the schema may be supplied by an old server

justheuristic:
from conversation with @artek0chumak : we found a better solution

We can use request.metadata (same for all `rpc_*`) to send field names

New client:
- send a varying number of tensors
- in request.metadata["tensor_names"] 
- for backward compatibility, always send (hidden_states, prompts, *optionally_more) to support old servers that expect 2 args
  - we can remove this compat code in 1.2.0

New server:
- in every handler, deserialize tensors and create a dict {name: tensor for name, tensor in zip(metadata["tensor_names"], tensors)}
- assert len(names) == len(tensors)
- backward compatibility: metadata.setdefault("tensor_names", ["hidden_states", "prompts"])

Old client:
- connects to new servers without sending tensor_names, but works due to the setdefault above

Old server:
- receives tensors with metadata, but ignores metadata
- if given (hidden_states, prompts), it ignores tensor_names but still works
- if given something else, fails with AssertionError - which is ok

justheuristic:
Note: if #225 is already merged, here's how you can update attention mask:
- that PR introduces no changes to rpc_forward/backward; the only meaningful change is with rpc_inference
- find `backend.py:_MergedInferenceStep.__call__`, add attention_mask as a parameter
    - inside __call__, forward attention mask to each block's rpc_inference
... and that's it!

borzunov:
#225 has been merged, feel free to proceed.

PR #227, opened 2023-01-19
Fix a typo in error message.

zsc:
By the code context, it can be inferred that do_sample==False when control reaches this point.


mryab (approved):
Thank you!

PR #232, opened 2023-01-23
Add instructions for updating to latest

vadi2:


borzunov:
Thanks for the PR! Right now, we recommend to use the latest stable Petals version from PyPI by default. Your case was an exception (I've made the server to choose `--num_blocks` for all models automatically only a few days ago, and I just didn't have time to ship a new stable version to PyPI - we plan to do it this week).

In turn, the server now automatically checks PyPI for updates and, if there is one, asks the user to run it and prints the corresponding command:



I think doing it this way is more intuitive than having a separate readme section, because users are unlikely to check for updates and the corresponding command manually. So if you don't mind, I'd keep it as is.

PR #243, opened 2023-01-30
Prompt-tuning notebooks: use latest Petals from the repo

borzunov:


borzunov:
I've eventually decided that we should just make a new PyPI release with the fix instead.

PR #251, opened 2023-02-05
Lower payload size threshold for stream handlers

borzunov:
Hotfix: we add "// 2" since hivemind==1.1.5 serializes bfloat16 tensors in float32, so they take 2x more space.

### DIFF
--- a/src/petals/client/remote_forward_backward.py
+++ b/src/petals/client/remote_forward_backward.py
@@ -108,7 +108,7 @@ async def run_remote_forward(
 
     # call RPC on remote server
     size = sum(t.element_size() * t.nelement() for t in inputs)
-    forward_fn = _forward_stream if size > MAX_UNARY_PAYLOAD_SIZE else _forward_unary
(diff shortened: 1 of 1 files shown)


src/petals/client/remote_forward_backward.py
```diff
@@ -108,7 +108,7 @@ async def run_remote_forward(
 
     # call RPC on remote server
     size = sum(t.element_size() * t.nelement() for t in inputs)
-    forward_fn = _forward_stream if size > MAX_UNARY_PAYLOAD_SIZE else _forward_unary
+    forward_fn = _forward_stream if size > MAX_UNARY_PAYLOAD_SIZE // 2 else _forward_unary
```
justheuristic, line 111: [nit] please check how MAX_UNARY_PAYLOAD_SIZE is defined - perhaps it's best to use MAX_MESSAGE_SIZE//4?

justheuristic (approved):

PR #252, opened 2023-02-05
Add bootstrap peer listening on port 80

borzunov:


borzunov:
Hmm, this won't actually help for firewalled machines because servers run on ports different from 80/443. I'll close this for now.

PR #286, opened 2023-03-11
Use number of tokens for attn_cache_size

mryab:
Previously, specifying the cache size in bytes was hard to reason about: the number of blocks is autotuned, so it's difficult to have a straightforward estimate for the optimal number of blocks vs number of tokens in the cache. This PR simplifies the server setup by providing an argument for the specific number of tokens to keep in the attention cache.

### DIFF
--- a/src/petals/cli/run_server.py
+++ b/src/petals/cli/run_server.py
@@ -4,9 +4,10 @@ import configargparse
 from hivemind.proto.runtime_pb2 import CompressionType
 from hivemind.utils.limits import increase_file_limit
 from hivemind.utils.logging import get_logger
+from humanfriendly import parse_size
 
 from petals.constants import PUBLIC_INITIAL_PEERS
-from petals.server.server import Server
+from petals.server.server import DTYPE_MAP, Server
 from petals.utils.version import validate_version
 
 logger = get_logger(__name__)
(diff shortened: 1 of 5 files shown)


src/petals/cli/run_server.py
```diff
@@ -78,11 +77,9 @@ def main():
     parser.add_argument("--torch_dtype", type=str, default="auto",
                         help="Use this dtype to store block weights and do computations. "
                              "By default, respect the dtypes in the pre-trained state dict.")
-    parser.add_argument('--attn_cache_size', type=str, default=None,
-                        help='The size of GPU memory allocated for storing past attention keys/values between inference steps. '
-                             'Examples: 500MB, 1.2GB, 1073741824 (bytes). Note that 1KB != 1KiB here. '
-                             'Default: 0.5GiB * num_blocks * hidden_size / 14336. '
-                             'The latter is the hidden size of the bigscience/bloom-petals model.')
+    parser.add_argument('--attn_cache_size', type=int, default=8192,
```
borzunov, line 80:
```suggestion
    parser.add_argument('--attn_cache_tokens', type=int, default=8192,
```

Let's rename it to **(1)** reflect that it's measured in tokens (in contrast to other options ending with `size`) and **(2)** avoid collisions (previously, int values were also correct and meant bytes - we need to show the error if someone used it).

borzunov (approved):

PR #292, opened 2023-03-15
Hotfix: Increase daemon_startup_timeout

borzunov:
For some reasons, right now 15 sec is not enough to connect to the bootstrap peers in the public swarm, as reported by multiple users and observed by me. Increasing it to 120 sec until we find the root cause of the issue.

### DIFF
--- a/src/petals/cli/run_server.py
+++ b/src/petals/cli/run_server.py
@@ -47,6 +47,9 @@ def main():
     parser.add_argument('--announce_maddrs', nargs='+', required=False,
                         help='Visible multiaddrs the host announces for external connections from other peers')
 
+    parser.add_argument('--daemon_startup_timeout', type=float, default=120,
+                        help='Timeout for the libp2p daemon connecting to initial peers')
+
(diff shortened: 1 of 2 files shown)


Likhith07:
Increasing it to 120 seconds and even 300 seconds doesnot solve this problem . kindly give me any other solution to this

borzunov:
Hi @Likhith07,

Sorry for taking ages to respond. The issue is likely to be not related to the timeout, but to the IPv6 config of your network (connection freezes on some configurations). I recommend to try fixing the configuration, or try providing the argument `--initial_peers /dns/bootstrap1.petals.ml/tcp/31337/p2p/QmedTaZXmULqwspJXz44SsPZyTNKxhnnFvYRajfH7MGhCY /dns/bootstrap2.petals.ml/tcp/31338/p2p/QmQGTqmM7NKjV6ggU1ZCap8zWiyKR89RViDXiqehSiCpY5`.

The latter is a subset of addresses of [public swarm's initial peers](https://github.com/bigscience-workshop/petals/blob/main/src/petals/constants.py#L1) that excludes IPv6 addresses.

PR #301, opened 2023-04-04
Start SequenceManager's thread only after first .make_sequence()

borzunov:
**Why?**

- We'd like to avoid excess threads for the original sequence manager in case if we only use its slices (e.g. when we add adapters or need only a subset of model blocks):

    ```python
    class BloomBasedClassifier(nn.Module):
      def __init__(self, model):
        super().__init__()
        self.distributed_layers = model.transformer.h
        self.adapter = nn.Sequential(nn.Linear(14336, 32), nn.Linear(32, 14336))
        self.head = nn.Sequential(nn.LayerNorm(14336), nn.Linear(14336, 2))
      
      def forward(self, embeddings):
        hidden_states = self.distributed_layers[0:6](embeddings)
        hidden_states = self.adapter(hidden_states)
        hidden_states = self.distributed_layers[6:10](hidden_states)
        pooled_states = torch.mean(hidden_states, dim=1)
        return self.head(pooled_states)
    ```

- If we create a sequence manager just before a fork (e.g. in a web app backend or a multi-thread benchmark), we'd like to avoid excess threads in the original process and only use this thread in child processes where we actually call `.make_sequence()`.

### DIFF
--- a/src/petals/client/remote_sequential.py
+++ b/src/petals/client/remote_sequential.py
@@ -48,7 +48,6 @@ class RemoteSequential(nn.Module):
                 request_timeout=config.request_timeout,
                 max_retries=config.max_retries,
                 allowed_servers=config.allowed_servers,
-                start=True,
                 **kwargs,
             )
             self.is_subsequence = False
--- a/src/petals/client/routing/sequence_manager.py
+++ b/src/petals/client/routing/sequence_manager.py
@@ -10,7 +10,7 @@ from typing import Any, Collection, Dict, List, Optional, Sequence, Union
 from weakref import WeakMethod
 
 import numpy as np
-from hivemind import DHT, P2P, MSGPackSerializer, PeerID
+from hivemind import DHT, P2P, MSGPackSerializer, PeerID, get_dht_time
 from hivemind.dht.node import Blacklist
 from hivemind.moe.client.remote_expert_worker import RemoteExpertWorker
 from hivemind.p2p import P2PHandlerError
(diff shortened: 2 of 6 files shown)


justheuristic (approved):
The code is okay, but there is a concern that we will wait longer when starting a colab notebook.

Suggestion:
- run 2 colab notebooks in each version; compare **time to first token generated**
- if difference is <10%, merge this freely
- if difference > 10%, we can trigger init in from_pretrained, while the weights are still loading

Please report your findings either way

borzunov:
@justheuristic I've benchmarked this short script:

```python
import transformers
from petals.client import DistributedBloomForCausalLM

model_name = "bigscience/bloom-petals"
tokenizer = transformers.BloomTokenizerFast.from_pretrained(model_name)
model = DistributedBloomForCausalLM.from_pretrained(model_name)

outputs = model.generate(max_new_tokens=1)
print(tokenizer.decode(outputs[0]))
```

Here are the results:

```python
In [2]: np.mean([22.144, 21.565, 21.270, 20.890, 20.632, 22.938, 21.354, 21.504, 21.619, 21.498])  # main
Out[2]: 21.5414

In [3]: np.mean([24.705, 24.246, 25.408, 23.942, 24.430, 24.595, 24.743, 24.137, 24.053, 23.166])  # lazy-seq-manager-start
Out[3]: 24.342499999999998

In [4]: (24.3425 - 21.5414) / 21.5414
Out[4]: 0.1300333311669623
```

The difference is 13%, so I'll go with the 2nd option. That is, I'll return the `start=True` argument but I'll make sure that it works even with `start=False` (in this case, it should create `SequenceManager()` in a lazy way).

PR #285, opened 2023-03-11
Speed up loading blocks using init with meta weights

mryab:
This PR changes the block initialization logic to use PyTorch meta tensors before assigning values from state_dict. This helps avoid unnecessary memory allocation and parameter initialization steps, which might take a lot of time (and RAM) for large models.

Before this PR: 1 block loaded in ~16 seconds
```
python -m petals.cli.run_server bigscience/bloom-petals --new_swarm --throughput 100 --num_blocks 1
Mar 11 20:28:06.177 [INFO] Running Petals 1.1.3
Mar 11 20:28:07.274 [INFO] This server is accessible directly
Mar 11 20:28:07.714 [INFO] Connecting to a private swarm, initial peers: []
Mar 11 20:28:07.715 [INFO] Running a server on ['/ip4/172.27.77.70/tcp/41059/p2p/12D3KooWA4oco1QGgLWRuJdVoyhCv1559B9x1tAjvomuKN8ZwYrY', '/ip4/127.0.0.1/tcp/41059/p2p/12D3KooWA4oco1QGgLWRuJdVoyhCv1559B9x1tAjvomuKN8ZwYrY', '/ip6/2a02:6b8:0:3201:d9cc:83ae:8057:2b4e/tcp/40189/p2p/12D3KooWA4oco1QGgLWRuJdVoyhCv1559B9x1tAjvomuKN8ZwYrY', '/ip6/::1/tcp/40189/p2p/12D3KooWA4oco1QGgLWRuJdVoyhCv1559B9x1tAjvomuKN8ZwYrY']
Mar 11 20:28:07.752 [INFO] Model weights will be loaded in 8-bit format
Mar 11 20:28:07.753 [INFO] Attention cache for all blocks will consume up to 0.50 GiB
Mar 11 20:28:07.843 [INFO] Reachability service started
Mar 11 20:28:11.264 [INFO] Announced that blocks [0] are joining
Mar 11 20:28:27.418 [INFO] Loaded bigscience/bloom-petals block 0, <All keys matched successfully>
```

After: 1 block loaded in ~2 seconds
```
python -m petals.cli.run_server bigscience/bloom-petals --new_swarm --throughput 100 --num_blocks 1
Mar 11 20:27:32.121 [INFO] Running Petals 1.1.3
Mar 11 20:27:33.257 [INFO] This server is accessible directly
Mar 11 20:27:33.700 [INFO] Connecting to a private swarm, initial peers: []
[...]

PR #310, opened 2023-04-25
Fix OOMs happening in case of accelerate >= 0.16.0

borzunov:
- After #285, `load_pretrained_block()` uses `accelerate.utils.set_module_tensor_to_device()`
- In accelerate>=0.16.0, it saves the tensor in the dtype previously used by the model instead of dtype of the weights (https://github.com/huggingface/accelerate/pull/920)
- Because of that, blocks and attention caches used float32, which caused OOMs
- This PR makes `load_pretrained_block()` respect `torch_dtype` (default: `"auto"`, which means reading `torch_dtype` from `config.json`)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ python_requires = >=3.7
 install_requires =
     torch>=1.12
     bitsandbytes==0.38.0.post2
-    accelerate>=0.15.0,<1.0.0
+    accelerate>=0.16.0,<1.0.0
     huggingface-hub>=0.11.1,<1.0.0
     transformers>=4.25.1,<5.0.0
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -33,7 +33,7 @@ python_requires = >=3.7
 install_requires =
     torch>=1.12
     bitsandbytes==0.38.0.post2
-    accelerate>=0.15.0,<1.0.0
+    accelerate>=0.16.0,<1.0.0
```
borzunov, line 36: `set_module_tensor_to_device`'s `dtype` arg didn't exist before 0.16.0

PR #319, opened 2023-05-25
Add benchmark scripts

borzunov:
This PR:

- Adds benchmark scripts for inference, forward pass, and full training step (e.g. used for experiments in our paper).
- Fixes bug with dtypes in `petals.DistributedBloomForSequenceClassification`.
- (minor refactor) Moves `DTYPE_MAP` to `petals.constants` as a useful constant.

### DIFF
--- a/benchmarks/benchmark_forward.py
+++ b/benchmarks/benchmark_forward.py
@@ -6,8 +6,9 @@ from time import perf_counter
 
 import torch
 from hivemind.utils.logging import get_logger
-from petals import DistributedBloomForCausalLM
 from transformers import BloomTokenizerFast
+
+from petals import DistributedBloomForCausalLM
 
 logger = get_logger()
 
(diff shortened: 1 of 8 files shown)


benchmarks/benchmark_forward.py
```diff
@@ -0,0 +1,65 @@
...
+def benchmark_forward(process_idx, args):
+    tokenizer = BloomTokenizerFast.from_pretrained(args.model)
+    model = DistributedBloomForCausalLM.from_pretrained(
+        args.model, initial_peers=args.initial_peers, torch_dtype=torch.bfloat16
+    )
+    logger.info(f"Created model: {process_idx=} {model.device=}")
+
+    torch.manual_seed(42)
+    for step in range(args.n_steps):
+        input_ids = torch.randint(100, 10000, size=(args.batch_size, args.seq_len))
```
justheuristic, line 48: [nit] we can use vocab size from config instead of 10k

benchmarks/benchmark_training.py
```diff
@@ -0,0 +1,95 @@
...
+    for step in range(args.n_steps):
+        input_ids = torch.randint(100, 10000, size=(args.batch_size, args.seq_len), device=args.device)
+        if args.task == "cls":
+            labels = torch.randint(0, 2, size=[args.batch_size], device=args.device)
+        else:
+            labels = input_ids
+
+        logger.info(f"{process_idx=} {step=} Forward")
+        start_time = perf_counter()
+        outputs = model(input_ids, labels=labels)
```
justheuristic, line 74: [nit] would be nice to explain the warmup batch, and, *optionally*, allow using multiple warmup batches

borzunov: [done]

benchmarks/benchmark_forward.py
borzunov: [done]

PR #323, opened 2023-06-08
Add LLaMA support

borzunov:
This PR:

1. **Abolishes the model conversion procedure.** Now, models are downloaded directly from original repositories like https://huggingface.co/bigscience/bloom. Servers download only shards with blocks to be hosted, and clients download only shards with input/output embeddings and layernorms.

    - BLOOM is loaded from `bigscience/bloom`, but we use the DHT prefix `bigscience/bloom-petals` for backward compatibility. Same with smaller BLOOMs and BLOOMZ.
    - LLaMA can be loaded from any repo like `username/llama-65b-hf`, but we use the DHT prefix `llama-65b-hf` (without the username) to accomodate blocks from different repos (there're a few of them with minor differences, such as `Llama` vs. `LLaMA` in the class name).

3. **Refactors the client to generalize it for multiple models.** Now, we have `petals.models` packages that contain model-specific code (e.g. `petals.models.bloom`, `petals.models.llama`). General code (e.g. CPU-efficient LM head, p-tuning) is kept in `petals.client`.

4. **Introduces** `WrappedLlamaBlock`, `DistributedLlamaConfig`, `DistributedLlamaForCausalLM`, `DistributedLlamaForSequenceClassification`, and `DistributedLlamaModel` compatible with Petals functionality (p-tuning, adapters, etc.).

5. **Introduces** `AutoDistributedConfig` that automatically chooses the correct config class (`DistributedLlamaConfig` or `DistributedBloomConfig`). The refactored configs contain all model-specific info for both clients and servers.

Upgrade instructions:

- Remove disk caches for blocks in old (converted) format to save disk space. That is, remove `~/.cache/petals/model--bigscience--bloom-petals` and  `~/.cache/petals/model--bigscience--bloomz-petals` directories (if present).

Tested:

- Servers hosting BLOOM and LLaMA
- Clients running inference, p-tuning and adapter tuning for BLOOM and LLaMA

Expected additions to this PR:

- [ ] Fix NaNs in prompt embeddings during LLaMA p-tuning
- [ ] Add `AutoDistributedModel`, `AutoDistributedModelForCausalLM`, `AutoDistributedModelForSequenceClassification` (so that we have Colab notebooks where it's enough to replace only the model name)

Future work for other PRs:

- [ ] Cover llama with tests
- [ ] Decide on cache reordering code
- [ ] Add guanaco
- [ ] Add falcon-40b and falcon-40b-instruct
- [ ] Update the "Host your own model" guide
- [ ] Update http://health.petals.ml and http://chat.petals.ml
- [ ] Upgrade example notebooks
- [ ] Add speed measurements vs. llama.cpp


.github/workflows/run-tests.yaml
```diff
@@ -80,8 +31,7 @@ jobs:
           pip install .[dev]
       - name: Test
         run: |
-          export HF_TAG=${{ hashFiles('setup.cfg', 'src/petals/cli/convert_model.py') }}
-          export MODEL_NAME=bloom-testing/test-bloomd-560m-$HF_TAG
+          export MODEL_NAME=bigscience/bloom-560m
```
borzunov, line 34: Loading the entire 560m model takes only 5 sec, so I'd stick to using it.

However, it takes a little more RAM, so we run 4 servers instead of 5 below.

.github/workflows/run-tests.yaml
```diff
@@ -6,57 +6,8 @@ on:
...
-          key: Key-v1-3.9-${{ hashFiles('setup.cfg') }}
-      - name: Install dependencies
-        if: steps.cache-model.outputs.cache-hit != 'true'
-        run: |
-          python -m pip install --upgrade pip
-          pip install .
-      - name: Delete any test models older than 1 week
-        if: steps.cache-model.outputs.cache-hit != 'true'
-        run: |
-          python tests/scripts/remove_old_models.py --author bloom-testing --use_auth_token $BLOOM_TESTING_WRITE_TOKEN
```
justheuristic, line 42: let's not forget to manually delete them in a week or so

src/petals/client/from_pretrained.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class FromPretrainedMixin:
+    @classmethod
+    def from_pretrained(
+        cls,
+        model_name_or_path: Union[str, os.PathLike, None],
+        *args,
+        low_cpu_mem_usage: bool = True,
+        torch_dtype: Union[str, torch.dtype] = "auto",
```
justheuristic, line 25: (nitpicking)
```suggestion
        torch_dtype: Optional[Union[str, torch.dtype]] = "auto",
```

src/petals/models/bloom/config.py
```diff
@@ -0,0 +1,37 @@
...
+
+logger = get_logger(__name__)
+
+
+class DistributedBloomConfig(BloomConfig, SequenceManagerConfig, PTuneConfig, LMHeadConfig):
+    block_class = WrappedBloomBlock
+    attn_class = BloomAttention
+    block_prefix = "h"
+
+    compat_suffix = "-prefix"
```
justheuristic, line 23: Nitpicking: can you please add a `#comment` why we need a compat_suffix

src/petals/models/llama/block.py
```diff
@@ -0,0 +1,87 @@
...
+        return (key_states, value_states)
+
+    def _reorder_cache_from_llama_to_bloom(
+        self, key_value: Tuple[torch.Tensor], batch_size: int, seq_length: int
+    ) -> Tuple[torch.Tensor]:
+        key_states, value_states = key_value
+        value_states = value_states.view(batch_size * self.self_attn.num_heads, seq_length, self.self_attn.head_dim)
+        key_states = key_states.view(*value_states.shape)
+        key_states = key_states.permute(0, 2, 1)
+        return (key_states, value_states)
```
justheuristic, line 87: Might be better to allow transformer blocks to define these methods than reordering from bloom.
If you agree, it's definitely okay to do it in a separate PR.
If you don't, please explain why.

src/petals/server/server.py
```diff
@@ -125,7 +131,7 @@ def __init__(
         self.dht = DHT(
             initial_peers=initial_peers,
             start=True,
-            num_workers=self.block_config.n_layer,
+            num_workers=self.block_config.num_hidden_layers,
```
justheuristic, line 134: sanity check: is this field guaranteed for all models or only for bloom and llama? couldn't find it

justheuristic (approved):
Wow, that's probably the largest petals PR of all time!

The new structure appears sound. The only real concern is that we hard-code bloom caching  order on the backend side -- instead of allowing each block to define their own cache order in as that block's methods or some other cunning plan. If you plan to do something similar, it's perfectly okay to do that later. If not, let's quickly discuss it, i might be missing something.

src/petals/server/server.py
borzunov: I think this field is standard because I saw models like BLOOM remapping model-specific config keys like `n_layer` to `num_hidden_layers` for compatibility (not vice versa): https://github.com/huggingface/transformers/blob/6ab045d6fe7a859ddc219cd144e638bb4d8ab2fe/src/transformers/models/bloom/configuration_bloom.py#L108

src/petals/models/bloom/config.py
borzunov: This was not actually used :) Removed it.

src/petals/client/from_pretrained.py
borzunov: Fixed

PR #325, opened 2023-06-10
Determine block dtype in a unified manner

mryab:


src/petals/server/block_utils.py
```diff
@@ -7,13 +7,13 @@
...
+def resolve_block_dtype(config: BloomConfig, dtype: Union[str, torch.dtype]) -> torch.dtype:
     """If dtype is "auto", resolves it using BloomConfig. Returns `dtype` intact otherwise."""
 
     if dtype == "auto" or dtype is None:
-        dtype = config.torch_dtype
-        if dtype == "auto" or dtype is None:
-            dtype = torch.float32
+        if config.torch_dtype == "auto" or config.torch_dtype is None:
+            return torch.bfloat16
+        return config.torch_dtype
```
borzunov, line 16: This nested ifs are difficult to understand, let's use linear code instead:

```python
if dtype not in ("auto", None):
    return dtype
if config.torch_dtype not in ("auto", None):
    return config.torch_dtype
return torch.bfloat16
```

tests/test_dtype.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+import torch
+
+from petals.bloom.from_pretrained import load_pretrained_block
+from petals.client import DistributedBloomConfig
+from petals.server.block_utils import resolve_block_dtype
+
+MODEL_NAME = "bloom-testing/test-bloomd-350m-main"
```
borzunov, line 8: Please use MODEL_NAME from test_utils, as in the rest of the tests

src/petals/server/server.py
```diff
@@ -174,14 +174,15 @@ def __init__(
                 first_block_index, last_block_index = map(int, map(str.strip, (first_block_index, last_block_index)))
             except Exception as e:
                 raise ValueError(f"Failed to parse `--block_indices {block_indices}`, must be start:end (e.g. 0:18)")
-            block_indices = range(first_block_index, last_block_index)
+            block_indices = list(range(first_block_index, last_block_index))
```
borzunov, line 177: Please leave `range()` here, so we have `range(0, 27)` in logs instead of a long line `[0, 1, ..., 26]`

borzunov (approved):

PR #335, opened 2023-06-30
Support peft LoRA adapters

artek0chumak:


src/petals/utils/peft.py
```diff
@@ -0,0 +1,102 @@
...
+    max_disk_space: Optional[int] = None,
+    delay: float = 30
+):
+    # TODO: Check is it possible to add safetensors loading inside petals/server/from_pretrained.py and reuse it here
+
+    if not check_peft_repository(repo_id):
+        raise ValueError(f"Repo: {repo_id} doesn't have safetensors inside for a safe loading.")
+    
+    try:
+        with allow_cache_reads(cache_dir):
```
borzunov, line 38: To keep code simpler, let's make a function with this code:

```python
def get_adapter_from_repo(repo_id, **kwargs):
      config_path = get_file_from_repo(repo_id, CONFIG_NAME, **kwargs)
      assert config_path is not None
      config = PeftConfig.from_json_file(config_path)
      weight_path = get_file_from_repo(repo_id, SAFETENSORS_WEIGHTS_NAME, **kwargs)
      assert weight_path is not None
      return config, load_file(weight_path)
```

src/petals/utils/peft.py
```diff
@@ -0,0 +1,102 @@
...
+        try:
+            with allow_cache_writes(cache_dir):
+                config_url = hf_hub_url(repo_id, CONFIG_NAME, revision=revision)
+                config_file_size = get_hf_file_metadata(config_url, token=use_auth_token).size
+                wieght_url = hf_hub_url(repo_id, SAFETENSORS_WEIGHTS_NAME, revision=revision)
+                wieght_file_size = get_hf_file_metadata(wieght_url, token=use_auth_token).size
+
+                file_size = config_file_size + wieght_file_size
+                if file_size is not None:
+                    free_disk_space_for(repo_id, file_size, cache_dir=cache_dir, max_disk_space=max_disk_space)
```
borzunov, line 73: Right now, `free_disk_space_for()` is only allowed to delete files from the same repository.

Could you please change it, so it is allowed to delete any files from `cache_dir`?

tests/test_peft.py
```diff
@@ -0,0 +1,49 @@
...
+import pytest
+import shutil
+
+from huggingface_hub import snapshot_download
+
+from petals.utils.peft import check_peft_repository, load_peft
+
+
+NOSAFE_PEFT_REPO = "timdettmers/guanaco-7b"
+SAFE_PEFT_REPO = "artek0chumak/guanaco-7b"
```
borzunov, line 11: Can you please replace this with a smaller pair of adapters? The current ones are too large to download in CI, this slows down tests by ~30%.

src/petals/utils/peft.py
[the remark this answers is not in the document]
artek0chumak: Thank you for the changes! Fix the issue in the next commit.

tests/test_peft.py
artek0chumak: Create smaller peft weights for `bloom-560m`

src/petals/utils/peft.py
artek0chumak: Thank you for the suggestions!

src/petals/utils/peft.py
```diff
@@ -0,0 +1,105 @@
...
+    is_tensors_found = dict()
+    with safe_open(filepath, framework=framework) as f:
+        for k in f.keys():
+            for layer_name in layers_name:
+                if k.startswith(layer_name):
+                    is_tensors_found[layer_name] = True
+                    tensors[k] = f.get_tensor(k)
+        for layer_name in layers_name:
+            if not is_tensors_found.get(layer_name, False):
+                logger.warning(f"There is no peft weights with prefix {layer_name}")
```
artek0chumak, line 33: I worry that the warning is not enough for this issue. Maybe I have to add an exception here?

src/petals/server/backend.py
```diff
@@ -80,6 +81,18 @@ def get_inference_cache_descriptors(self, batch_size: int, max_length: int) -> S
             cache_tensors.extend((keys, values))
         return cache_tensors
 
+    def forward(self, *inputs: Union[torch.Tensor, str]) -> Tuple[torch.Tensor, ...]:
+        *inputs, active_adapter = inputs
+        if active_adapter and not self.load_adapter_(active_adapter):
+            raise KeyError("Could not find adapter {inference_info.active_adapter}; perhaps it is not loaded")
```
borzunov, line 87:
```suggestion
            raise KeyError(f"Could not find adapter {inference_info.active_adapter}, perhaps it is not loaded")
```

src/petals/server/backend.py
```diff
@@ -80,6 +81,18 @@ def get_inference_cache_descriptors(self, batch_size: int, max_length: int) -> S
...
+        return super().forward(*inputs)
+
+    def backward(self, *inputs: Union[torch.Tensor, str]) -> Tuple[torch.Tensor, ...]:
+        *inputs, active_adapter = inputs
+        if active_adapter and not self.load_adapter_(active_adapter):
+            raise KeyError("Could not find adapter {inference_info.active_adapter}; perhaps it is not loaded")
```
borzunov, line 93:
```suggestion
            raise KeyError(f"Could not find adapter {inference_info.active_adapter}, perhaps it is not loaded")
```

src/petals/client/remote_sequential.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         dht: Optional[DHT] = None,
         start_block: Optional[int] = None,
         end_block: Optional[int] = None,
+        **kwargs,
```
justheuristic, line 31: this is a hook to pass things like adapter ID;

nb: the **kwargs are not passed infinitely; any args that are not explicitly declared in RemoteSequenceManager.__init__ will cause an error.

src/petals/dht_utils.py
```diff
@@ -105,7 +121,12 @@ async def _get_remote_module_infos(
         for peer_id, server_info in metadata.value.items():
             try:
                 peer_id = PeerID.from_base58(peer_id)
-                state, throughput = server_info.value
+                state, throughput = server_info.value[:2]
+                available_adapters = server_info.value[2] if len(server_info.value) > 2 else []
```
justheuristic, line 125: translation: if old server then no adapters

justheuristic:
## Comments:

1. It seems that the [test adapter](https://huggingface.co/artek0chumak/bloom-560m-safe-peft/tree/main) is initialized at zeros; therefore, the test for exact match with adapters does not fully test exact match.
   - can you please update safe peft to make sure that they affect outputs in a measurable way?
   - we can use any dummy weights s.t. not allclose(outputs_with_peft, outputs_without_peft)

2. Currently we are using `tensor_parallel` with adapters naively, as though it was non-parallel
  - this may cause problems in case when adapters are non-zero
  - unless it is super-easy to fix, consider simply `assert not adapters or not tensor_parallel` for now
  - naive, but correct solution for later: initialize shards[0] with correct adapters; shards[1,2,...] with no adapters
  - optimal, but complicated solution for later: split adapter evenly along the low-rank dimension; this can be done using tensor_parallel's config

tests/test_full_model.py
```diff
@@ -12,12 +13,16 @@
...
+@pytest.mark.parametrize("use_peft", (True, False) if ADAPTER_NAME else (False,))
 @pytest.mark.parametrize("pass_empty_tensors", (True, False))
-def test_full_model_exact_match(pass_empty_tensors: bool, atol_forward=1e-3, atol_inference=1e-3):
+def test_full_model_exact_match(use_peft: bool, pass_empty_tensors: bool, atol_forward=1e-3, atol_inference=1e-3):
     tokenizer = transformers.BloomTokenizerFast.from_pretrained(MODEL_NAME)
     model = DistributedBloomForCausalLM.from_pretrained(
         MODEL_NAME, initial_peers=INITIAL_PEERS, low_cpu_mem_usage=True, torch_dtype=torch.float32
     )
+    if use_peft:
+        model.transformer.h.sequence_manager.extra_metadata = dict(active_adapter=ADAPTER_NAME)
```
justheuristic, line 24: Note: ideally, sequence manager should be created with the correct adapter; otherwise, it will perform the first pass over servers without that adapter

artek0chumak:
> 1. It seems that the [test adapter](https://huggingface.co/artek0chumak/bloom-560m-safe-peft/tree/main) is initialized at zeros; therefore, the test for exact match with adapters does not fully test exact match.

Thank you for mentioning this! Change LoRA weights in hub, currently they have non-zero parameters.

Issue #327, opened 2023-06-13
Prompt Tuning met NaN error


xinghua-qu:
I tried to run the code for [prompt tuning example](https://github.com/bigscience-workshop/petals/blob/main/examples/prompt-tuning-personachat.ipynb), but got the NaN error after some iterations.

Does anybody know the reason for such error? Thanks :)
The dots after the training curve mean NaN.
<img width="499" alt="image" src="https://github.com/bigscience-workshop/petals/assets/36146785/74e22a3e-c57b-408d-a55a-035d56c6c011">

Issue #283, opened 2023-03-10
Exploding gradients/weights/logits in a fine-tuning notebook


81549361:
Hello, after I successfully fine-tuned, I run to the last step and when generating text, the following error occurs. 
May I ask what is the reason?

![image](https://user-images.githubusercontent.com/7599454/224216132-416b7044-dee1-4401-a7c7-c2f76ea37660.png)
![image](https://user-images.githubusercontent.com/7599454/224216173-81b75b67-79d0-4b6f-a654-2cf500ea07d5.png)

borzunov:
Hi @81549361, sorry for the long response.

One possible reasons is that model weights/gradients have exploded during training. You would see this on the loss function plot in wandb.

We'll double-check training hyperparams and fix them to avoid this behavior (cc @justheuristic).

mryab:
Hi @81549361, we've just pushed a commit that should fix the explosion in https://github.com/bigscience-workshop/petals/pull/343. Can you try running the updated SST-2 prompt tuning notebook from the main branch?

Issue #241, opened 2023-01-29
The prompt tuning example (prompt-tuning-sst2) don't work


Gad1001:
Hi,
I have tried to run the notebook example you have published (without editing) in colab but it doesn't work...
I get the following error:
```
RuntimeError                              Traceback (most recent call last)
[<ipython-input-12-7f7d7fa267e9>](https://localhost:8080/#) in <module>
     17 
     18         model.train()
---> 19         outputs = model(**batch)
     20         loss = outputs.loss
     21         loss.backward()

3 frames
[/usr/local/lib/python3.8/dist-packages/torch/nn/modules/linear.py](https://localhost:8080/#) in forward(self, input)
    112 
    113     def forward(self, input: Tensor) -> Tensor:
--> 114         return F.linear(input, self.weight, self.bias)
    115 
    116     def extra_repr(self) -> str:

RuntimeError: expected scalar type Half but found Float
```
It is probably about the transformers version out of date but petals 1.1.1 requires transformers==4.25.1

justheuristic:
@artek0chumak can you please check?

justheuristic:
While @artek0chumak  is on his way, you might want a more basic example here: 
https://github.com/bigscience-workshop/petals/issues/233

Just paste the code in the basic notebook

artek0chumak:
Hi!

This issue was fixed in the repo, but it's still not in the "pip version".

Please install petails via:
pip install https://github.com/bigscience-workshop/petals/archive/main.zip

... and the issue should be resolved. If not, please tell us.

borzunov:
I have fixed this in #243. Also, I'm going to make a new release soon.

borzunov:
I have just shipped the new PyPI release (v1.1.2), the notebooks should now work as is.

Gad1001:
Hi, the issue has been solved, but now another error appears:
```
AttributeError: module 'torch.nn' has no attribute 'CrossEntoryCriterion'

     12 lr_scheduler = get_scheduler(
     11 
---> 10 cls_criterion = nn.CrossEntoryCriterion()
      9 cls_optimizer = AdamW(cls_model.parameters(), lr=LR, weight_decay=WEIGHT_DECAY)
      8 )
[<ipython-input-13-c8610e78a6af>](https://localhost:8080/#) in <module>
AttributeError                            Traceback (most recent call last)
---------------------------------------------------------------------------
```

justheuristic:
[...]

PR #343, opened 2023-07-12
Fix convergence issues and switch to LLaMA in the SST-2 example

mryab:
This PR switches the model to LLaMA in the SST-2 example notebook, as well as fixes NaNs appearing during training by using proper mixed precision with loss scaling. I've removed deep p-tuning for now for the sake of simplicity, will restore and fix the PersonaChat notebook once I fix it as well.

The example run is here: https://wandb.ai/mryab/bloom-sst-2/runs/3e0suayd?workspace=user-mryab. As we can see, the loss is still a bit unstable, bug at least we have no outright divergence:
image

When merged, this should fix https://github.com/bigscience-workshop/petals/issues/327, https://github.com/bigscience-workshop/petals/issues/283 and https://github.com/bigscience-workshop/petals/issues/241.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,3 +126,5 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+.idea/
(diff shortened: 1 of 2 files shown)


justheuristic (approved):
LGTM

poedator:
> The example run is here: https://wandb.ai/mryab/bloom-sst-2/runs/3e0suayd?workspace=user-mryab
this link is not valid. 404

mryab:
Just made it public, thanks for noticing!

PR #345, opened 2023-07-12
Import petals.utils.peft only when needed to avoid unnecessary import of bitsandbytes

borzunov:
The motivation is the same as in #180.


justheuristic (approved):
Definitely LGTM; I have a sneaking suspicion that we've probably imported peft elsewhere as well.

PR #346, opened 2023-07-12
Estimate adapter memory overhead in choose_num_blocks()

justheuristic:
This PR changes the way the server determines how many blocks it can handle if that server has at least one adapter. It does so by including the adapter parameters in the total block size that divides the free GPU memory.

ToDos:
- [ ] make sure adapters are actually loaded in the same dtype that was used
- [ ] update artek0chumak/bloom-560m-safe-peft to disable dropout
- [ ] create github issues from #335 comments

### DIFF
--- a/src/petals/server/server.py
+++ b/src/petals/server/server.py
@@ -253,7 +253,7 @@ class Server:
         if adapters:
             # Delay import of petals.utils.peft to avoid unnecessary import of bitsandbytes
             from petals.utils.peft import estimate_adapter_memory_per_block
-            
+
             adapter_memory_per_block = estimate_adapter_memory_per_block(
                 self.block_config, self.torch_dtype, self.adapters, self.cache_dir
             )
--- a/src/petals/server/server.py
+++ b/src/petals/server/server.py
@@ -30,6 +30,7 @@ from petals.server.throughput import get_dtype_name, get_server_throughput
 from petals.utils.auto_config import AutoDistributedConfig
 from petals.utils.convert_block import QuantType, check_device_balance, convert_block
 from petals.utils.disk_cache import DEFAULT_CACHE_DIR
+from petals.utils.peft import estimate_adapter_memory_per_block
 from petals.utils.version import get_compatible_model_repo
 
 logger = get_logger(__name__)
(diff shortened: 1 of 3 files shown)


src/petals/utils/peft.py
```diff
@@ -194,15 +200,35 @@ def add_adapter_to_block(block, block_index, adapter_name, peft_config, peft_sta
...
+        create_lora_adapter(block, quant_type=QuantType.NONE)
+
+        for adapter in adapters:
+            peft_config, peft_state_dict = load_peft(adapter, block_idx=0, **kwargs)
+            assert peft_config["peft_type"].upper() == "LORA", "only LoRA adapters are supported for now"
+            add_adapter_to_block(
+                block, block_index=0, adapter_name=adapter, peft_config=peft_config, peft_state_dict=peft_state_dict
+            )
+        adapter_parameters = sum(p.numel() for p in block.parameters()) - base_block_parameters
+    bytes_per_parameter = torch.finfo(resolve_block_dtype(block_config, torch_dtype)).bits / 8
```
borzunov, line 233: Doesn't adapters have a different dtype from the base block?

src/petals/server/server.py
```diff
@@ -250,7 +250,12 @@ def _choose_num_blocks(self) -> int:
         # Estimate of GPU memory used in rpc_backward (2 GiB for BLOOM, proportional for other models)
         autograd_memory = 2 * gib * num_devices / 14336 * self.block_config.hidden_size
 
-        num_blocks = math.floor((total_memory - autograd_memory) / (block_size + self._cache_bytes_per_block))
+        adapter_memory_per_block = estimate_adapter_memory_per_block(
+            self.block_config, self.torch_dtype, self.adapters, self.cache_dir
+        )
```
borzunov, line 255:
```suggestion
        if adapters:
            # Delay import of petals.utils.peft to avoid unnecessary import of bitsandbytes
            from petals.utils.peft import estimate_adapter_memory_per_block
            
            adapter_memory_per_block = estimate_adapter_memory_per_block(
                self.block_config, self.torch_dtype, self.adapters, self.cache_dir
            )
```

src/petals/utils/peft.py
```diff
@@ -1,9 +1,13 @@
 import re
 import time
-from typing import List, Optional
+from typing import List, Optional, Sequence
```
borzunov, line 3:
```suggestion
import os
from typing import List, Optional, Sequence

os.environ["BITSANDBYTES_NOWELCOME"] = "1"
```

borzunov: Since we do it here, please remove this line at `petals/utils/convert_block.py:59`.

src/petals/utils/peft.py
justheuristic: they should have the same, lest forward pass would cause an error

PR #351, opened 2023-07-14
Test that bitsandbytes is not imported when it's not used

borzunov:
We avoid importing bitsandbytes when it's not used, since bitsandbytes doesn't always find correct CUDA libs and may raise exceptions because of that.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ python_requires = >=3.7
 install_requires =
     torch>=1.12
     bitsandbytes==0.40.0.post4
-    accelerate>=0.16.0,<1.0.0
+    accelerate>=0.16.0,<0.21.0
     huggingface-hub>=0.11.1,<1.0.0
     tokenizers>=0.13.3
     transformers>=4.30.1,<5.0.0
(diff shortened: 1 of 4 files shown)


tests/test_sequence_manager.py
```diff
@@ -43,7 +43,7 @@ def test_sequence_manager_basics(mode: str):
     assert shutdown_evt.is_set()
 
 
-class TestSequenceManager(RemoteSequenceManager):
+class RemoteSequenceManagerWithChecks(RemoteSequenceManager):
```
borzunov, line 46: pytest thinks that classes whose name starts with `Test` are test cases and raises warnings during test collection. Fixing that.

artek0chumak (approved):

setup.cfg
```diff
@@ -33,7 +33,7 @@ python_requires = >=3.7
 install_requires =
     torch>=1.12
     bitsandbytes==0.40.0.post4
-    accelerate>=0.16.0,<1.0.0
+    accelerate>=0.16.0,<0.21.0
```
borzunov, line 36: accelerate 0.21.0 released yesterday imports bnb automatically if it's installed :( Maybe this will change in future.

PR #353, opened 2023-07-14
Switch adapters slightly faster

justheuristic:
Currently, each `TransformerBackend.inference_step` looks for adapters and sets the correct adapter type for each block. This is not very expensive, but it can measurably affect inference time:

image

This pull request uses faster adapter switching with just one variable assignment, without iterating over block.modules().

### DIFF
--- a/src/petals/server/handler.py
+++ b/src/petals/server/handler.py
@@ -68,6 +68,7 @@ class TransformerConnectionHandler(ConnectionHandler):
         dht: DHT,
         module_backends: Dict[str, TransformerBackend],
         *,
+        adapters: Optional[Sequence[str]],
         dht_prefix: str,
         push_manager: multiprocessing.managers.SyncManager,
         session_queues: Dict[str, multiprocessing.managers.BaseProxy],  # BaseProxy for queue.Queue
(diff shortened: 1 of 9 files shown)


src/petals/server/handler.py
```diff
@@ -142,6 +144,7 @@ async def rpc_inference(
                 requested_backends = tuple(self.module_backends[uid] for uid in requested_uids)
                 max_length = metadata.get("max_length")
                 active_adapter = metadata.get("active_adapter", "")
+                assert not active_adapter or active_adapter in self.adapters
```
borzunov, line 147: Please add a error message here, otherwise the client will receive P2PHandlerError("") and won't know what happened

src/petals/utils/peft.py
```diff
@@ -118,6 +119,47 @@ def load_peft(
             time.sleep(delay)
 
 
+class GlobalAdapterMixin:
+    """A mixin that makes LoRA-wrapped linear layers obey a globally set adapter"""
+
+    ADAPTER_NOT_SET = "__ADAPTER_NOT_SET"
+    GLOBAL_ACTIVE_ADAPTER = ADAPTER_NOT_SET
```
borzunov, line 126:
```suggestion
    global_active_adapter = None
```

1. `GLOBAL_ACTIVE_ADAPTER` is not a constant, it should not be in caps
2. Why use a predefined string constant instead of just `None`?

borzunov (approved):

PR #354, opened 2023-07-14
Fix bugs in _choose_num_blocks() added in #346

borzunov:
Fix bugs from #346.

### DIFF
--- a/src/petals/server/server.py
+++ b/src/petals/server/server.py
@@ -174,9 +174,13 @@ class Server:
         self.quant_type = quant_type
         logger.info(f"Model weights are loaded in {get_dtype_name(torch_dtype, quant_type)} format")
 
+        # For attention cache in GPU or RAM
         cache_values_per_block = 2 * self.block_config.hidden_size * attn_cache_tokens
         self._cache_bytes_per_block = cache_values_per_block * torch.finfo(self.torch_dtype).bits // 8
+
+        # For disk cache
         self.cache_dir = cache_dir
+        self.max_disk_space = max_disk_space
         self.adapters = adapters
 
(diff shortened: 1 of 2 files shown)


src/petals/server/server.py
```diff
@@ -249,12 +250,17 @@ def _choose_num_blocks(self) -> int:
         # Estimate of GPU memory used in rpc_backward (2 GiB for BLOOM, proportional for other models)
         autograd_memory = 2 * gib * num_devices / 14336 * self.block_config.hidden_size
 
-        if adapters:
+        if self.adapters:
```
borzunov, line 253: `adapters` was not defined

src/petals/server/server.py
```diff
@@ -249,12 +250,17 @@ def _choose_num_blocks(self) -> int:
         # Estimate of GPU memory used in rpc_backward (2 GiB for BLOOM, proportional for other models)
         autograd_memory = 2 * gib * num_devices / 14336 * self.block_config.hidden_size
 
-        if adapters:
```
borzunov, line 252: `adapters` is not defined

src/petals/server/server.py
```diff
@@ -249,12 +250,17 @@ def _choose_num_blocks(self) -> int:
...
+        if self.adapters:
             # Delay import of petals.utils.peft to avoid unnecessary import of bitsandbytes
             from petals.utils.peft import estimate_adapter_memory_per_block
 
             adapter_memory_per_block = estimate_adapter_memory_per_block(
-                self.block_config, self.torch_dtype, self.adapters, self.cache_dir
```
borzunov, line 257: `cache_dir` is a kwarg-only. Also, we need all the rest of disk cache args

PR #358, opened 2023-07-16
Report inference, forward, and network RPS separately

borzunov:
Inference RPS may be very different from forward RPS. E.g., currently bnb uses a completely different algorithm for NF4 inference. We report detailed RPS info that can be then used for shortest-path routing for inference.

### DIFF
--- a/src/petals/utils/ping.py
+++ b/src/petals/utils/ping.py
@@ -16,7 +16,7 @@ async def ping(
     _dht: hivemind.DHT,
     node: hivemind.dht.DHTNode,
     *,
-    wait_timeout: float = 1,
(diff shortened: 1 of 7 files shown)


src/petals/utils/ping.py
```diff
@@ -16,7 +16,7 @@ async def ping(
     _dht: hivemind.DHT,
     node: hivemind.dht.DHTNode,
     *,
-    wait_timeout: float = 1,
+    wait_timeout: float = 5,
```
borzunov, line 19: Peers that use relay may yield large first ping, but it'll be smoothed in future.

PR #356, opened 2023-07-15
Make a server ping next servers

borzunov:
This PR makes a server ping potential next servers in a chain and report the RTTs to DHT:

Screenshot 2023-07-15 at 20 14 55

This is going to be used for shortest-path routing.

PR #362, opened 2023-07-17
Implement shortest-path routing for inference

borzunov:
This PR:

1. **Adds shortest path routing for inference.** We build a graph with client-server and server-server latencies and compute costs, as well as empirically measured overheads. For client-server latencies, we ping possible first and last servers in a sequence in `SequenceManager.update()`. We penalize servers who may not have enough cache for our request. This uses info added to DHT in #355, #356, #358.

2. **Makes a server ping neighboring servers in addition to next ones.** This is to get an opportunity to change the server even before we use all its blocks (e.g., because a neighboring server is faster). This feature is not enabled though, since it increases graph size for N servers to O(N^2) - but we may enable it if needed.

3. **Fixes a `SequenceManager` bug with the first `update()`.** Previously, this update was likely to produce incorrect information and cause to `MissingBlocksErrors` until the next update happens.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,6 +48,7 @@ install_requires =
     sentencepiece>=0.1.99
     peft@git+https://github.com/huggingface/peft@5884bdbea49e5e71e2cd06ecfa484bb635063735
     safetensors>=0.3.1
+    Dijkstar>=2.6.0
 
 [options.extras_require]
 dev =
--- a/src/petals/client/routing/sequence_manager.py
+++ b/src/petals/client/routing/sequence_manager.py
@@ -212,6 +212,8 @@ class RemoteSequenceManager:
         end_index: int,
         *,
         cache_tokens_needed: Optional[int],
+        overhead_coeff: float = 1.82,  # Backend overhead (empirically measured)
+        overhead_delay: float = 0.018,  # Serialization overhead (empirically measured)
         default_inference_rps: float = 300,  # If inference RPS unknown
         alloc_delay: float = 60,  # If not enough cache left, we penalize the edge
     ) -> dijkstar.Graph:
@@ -230,6 +232,7 @@ class RemoteSequenceManager:
         # Clent -> server network delays
         for span in self.state.sequence_info.spans_containing_block[start_index]:
             delay = self._rtt_to_delay(client_server_rtts.get(span.peer_id))
+            delay += overhead_delay
             if not self._has_cache_for(span, cache_tokens_needed):
                 delay += alloc_delay
             graph.add_edge("start", (span.peer_id, start_index), delay)
@@ -252,6 +255,7 @@ class RemoteSequenceManager:
                     if cur_span.server_info.next_pings is not None:
                         rtt = cur_span.server_info.next_pings.get(next_span.peer_id.to_base58())
                     delay = self._rtt_to_delay(rtt)
+                    delay += overhead_delay
                     if not self._has_cache_for(next_span, cache_tokens_needed):
                         delay += alloc_delay
                     graph.add_edge((cur_span.peer_id, block_idx), (next_span.peer_id, block_idx), delay)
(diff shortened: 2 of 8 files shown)


setup.cfg
```diff
@@ -48,6 +48,7 @@ install_requires =
     sentencepiece>=0.1.99
     peft@git+https://github.com/huggingface/peft@5884bdbea49e5e71e2cd06ecfa484bb635063735
     safetensors>=0.3.1
+    Dijkstar>=2.6.0
```
justheuristic, line 51: nit: perhaps we should set <3.0.0 to protect against interface changes

src/petals/client/routing/sequence_manager.py
```diff
@@ -143,27 +156,152 @@ def make_sequence(
...
+
+        return span_sequence
+
+    def _build_inference_graph(
+        self,
+        start_index: int,
+        end_index: int,
+        *,
+        cache_tokens_needed: Optional[int],
+        default_inference_rps: float = 300,  # If inference RPS unknown
```
justheuristic, line 218: nit: 300 layer-tokens sounds a bit optimistic for unknown GPUs

src/petals/client/routing/sequence_manager.py
```diff
@@ -143,27 +156,152 @@ def make_sequence(
...
+        missing_blocks = [
+            block_idx
+            for block_idx in range(start_index, end_index)
+            if not self.state.sequence_info.spans_containing_block[block_idx]
+        ]
+        if missing_blocks:
+            raise MissingBlocksError(missing_blocks)
+
+        client_server_rtts = self.ping_aggregator.to_dict()
+        logger.info(f"Client-server RTTs: {client_server_rtts}")
```
justheuristic, line 230: Nit: do we want to print it to every **client**? If not, consider logger.debug

src/petals/client/routing/sequence_manager.py
```diff
@@ -143,27 +156,152 @@ def make_sequence(
...
+                        delay += alloc_delay
+                    graph.add_edge((cur_span.peer_id, block_idx), (next_span.peer_id, block_idx), delay)
+
+        # Compute delays
+        for span in self.state.sequence_info.spans_by_priority:
+            for block_idx in range(max(span.start, start_index), min(span.end, end_index)):
+                inference_rps = span.server_info.inference_rps
+                if inference_rps is None:
+                    inference_rps = default_inference_rps
+                graph.add_edge((span.peer_id, block_idx), (span.peer_id, block_idx + 1), 1 / inference_rps)
```
justheuristic, line 267: nit: do you, by chance, have some mockup performance numbers for graph construction & pathfinding for simulated larger graphs? If no, please remind me to run them.

justheuristic (approved):

src/petals/client/routing/sequence_manager.py
```diff
@@ -143,27 +156,152 @@ def make_sequence(
...
+        return min(rtt / 2, max_delay)
+
+    @staticmethod
+    def _has_cache_for(span: RemoteSpanInfo, cache_tokens_needed: Optional[int] = None) -> bool:
+        if cache_tokens_needed is None or span.server_info.cache_tokens_left is None:
+            return True
+
+        # This is a pessimistic estimate that assumes that we'll use all blocks hosted by this server,
+        # which is not always true. This is okay since false positives are more costly than false negatives here.
+        return cache_tokens_needed * 2 * span.length <= span.server_info.cache_tokens_left
```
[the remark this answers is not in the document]
borzunov: It's already reported this way. I've updated the comment to improve clarity.

setup.cfg
borzunov: Here, I use only the simplest graph interface - I think it's unlikely to be changed and/or can be quickly updated.

src/petals/client/routing/sequence_manager.py
borzunov: These are really low-level constants that's unlikely to be changed by a user, so I wouldn't pollute the config with them (it already has lots of knobs).

src/petals/client/routing/sequence_manager.py
borzunov: Sure! I use them for debug, I'll convert all `logger.info()` to `logger.debug()` before merging

src/petals/client/routing/sequence_manager.py
borzunov: If we allow switching a server before going to the end of the first one, we get O(N^2) graph for servers. For N = 50 servers holding blocks 0..80, building a graph and looking for the shortest path takes ~0.3 sec and scales as O(N^2).

I'll forbid such switching, this makes graph much smaller (takes ~0.03 sec now) and scales much better.

PR #377, opened 2023-07-19
Inherit bitsandbytes compute dtype correctly (override peft quirk)

justheuristic:
see https://github.com/huggingface/peft/pull/732


borzunov:
### Chatbot speeds

Before fix:

guanaco: 3.6-3.7 tok/s
llama: 3.7-3.9 tok/s

After fix:

guanaco: 4.7 tok/s
llama: 6.5 tok/s

PR #380, opened 2023-07-19
Update model.py

mzamini92:
Isn't it better to use jit decorator to improve the performance? The torch.jit.script_method decorator will compile the forward pass of the model into a Torch script. This will make the forward pass faster and more efficient to execute. The torch.jit.script_method decorators are correctly applied to the forward, _forward_tune_mode, and _forward_default_mode methods of DistributedLlamaModel, ensuring that they are compiled as Torch scripts. The rest of the code remains unchanged.

### DIFF
--- a/src/petals/models/llama/model.py
+++ b/src/petals/models/llama/model.py
@@ -36,6 +36,7 @@ class DistributedLlamaModel(FromPretrainedMixin, PTuneMixin, LlamaModel):
         self.requires_grad_(False)  # Forbid accumulate grads for embeddings and layernorm
         self.init_prompts(config)
 
+    @torch.jit.script_method
     def forward(
         self,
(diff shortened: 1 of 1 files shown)


borzunov:
Hi @mzamini92,

Thanks for the PR.

I'm afraid your code does not have any effect because transformer block computations are done remotely on servers. That is, there are almost no torch computations inside functions you've wrapped in `torch.jit.script`.

There is a chance that something like `torch.jit.script` may speed up server computations - in this case, you'd need to add that to `petals.server` package code.

If you'd like to try, it is important to conduct benchmarks and actually show that inference/fine-tuning becomes faster with your changes (the server code is already optimized, and some changes may actually slow it down). We won't merge any optimizations without proper benchmarks conducted for different models and network conditions.

PR #397, opened 2023-07-21
Update README.md to explain how to use older python versions

SephReed:
python 11 has issues


borzunov:
Fixed in #393, Python 3.11 now works.

PR #403, opened 2023-07-22
Split long sequences into chunks

justheuristic:
ToDo
- [ ] make similar stuff for forward and backward


src/petals/server/server.py
```diff
@@ -58,6 +58,7 @@ def __init__(
         inference_max_length: Optional[int] = None,
         min_batch_size: int = 1,
         max_batch_size: Optional[int] = None,
+        max_chunk_size_bytes: int = 1024,
```
borzunov, line 61:
```suggestion
        max_chunk_size_bytes: int = 512 * 1024 * 1024,
```

src/petals/server/backend.py
```diff
@@ -105,14 +114,39 @@ def inference_step(
...
+                    hidden_states_chunk, layer_past=layer_past, use_cache=True
+                )
+                if num_chunks > 1:
+                    output_hidden_states[
+                        :, chunk_i * max_chunk_length : (chunk_i + 1) * max_chunk_length, :
+                    ] = output_hidden_states_chunk
+                else:
+                    output_hidden_states = output_hidden_states_chunk  # saves one memcopy
+                layer_past = new_kvs
+            del layer_past
```
borzunov, line 140:
```suggestion
```

Why `del` if we return right away?

PR #428, opened 2023-08-02
Penalize servers that use relays during rebalancing

borzunov:
Servers accessible only via relays may introduce issues if they are the only type of servers holding certain blocks. Specifically, a connection to such servers may be unstable or opened after a certain delay.

This PR changes their self-reported throughput, so that the rebalancing algorithm prefers to put directly available servers for hosting each block.

### DIFF
--- a/src/petals/server/server.py
+++ b/src/petals/server/server.py
@@ -83,7 +83,7 @@ class Server:
         quant_type: Optional[QuantType] = None,
         tensor_parallel_devices: Optional[Sequence[torch.device]] = None,
         skip_reachability_check: bool = False,
-        dht_client_mode: Optional[bool] = None,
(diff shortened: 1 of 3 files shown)


src/petals/server/server.py
```diff
@@ -83,7 +83,7 @@ def __init__(
         quant_type: Optional[QuantType] = None,
         tensor_parallel_devices: Optional[Sequence[torch.device]] = None,
         skip_reachability_check: bool = False,
-        dht_client_mode: Optional[bool] = None,
+        reachable_via_relay: Optional[bool] = None,
```
borzunov, line 86: Refactor: Using a proper name instead

justheuristic (approved):
LGTM, looks like this will indeed alleviate inference speed issues if there is only one relay behind a block

PR #434, opened 2023-08-04
Rewrite MemoryCache alloc_timeout logic

justheuristic:
- rpc_inference: server will now accept allocation timeout from user, defaults to no timeout
- bugfix: inference timeout is now measured from the moment the request is received
   - previously, you would have to wait for your timeout plus the time it takes to sort through the queue (other users' timeout)
   - now, you get AllocationFailed if you had to wait for over (timeout) seconds - regardless of other users
- a request for inference with no timeout will now fail instantly if there is not enough memory available
- dtype number of bytes is now correctly determined for int, bool & other types

**Testing:**

- [x] test that nonzero timeout includes time spent waiting for lock
- [x] test that zero timeout is always instantaneous
- [x] test that zero timeout always works if cache has free memory
- [x] test integration with client's InferenceSession


src/petals/cli/run_server.py
```diff
@@ -96,9 +96,9 @@ def main():
     parser.add_argument("--torch_dtype", type=str, choices=DTYPE_MAP.keys(), default="auto",
                         help="Use this dtype to store block weights and do computations. "
                              "By default, respect the dtypes in the pre-trained state dict.")
-    parser.add_argument('--alloc_timeout', type=float, default=1,
-                        help='If the cache is full, the server will wait for this number of seconds hoping that some memory will be freed '
-                             'before rejecting the request')
+    parser.add_argument('--max_alloc_timeout', type=float, default=600,
```
justheuristic, line 99: nb: large max timeout is no longer a problem because long-timeout users will no longer cause others to freeze

tests/test_cache.py
```diff
@@ -0,0 +1,172 @@
...
+    descr_c = TensorDescriptor.from_tensor(torch.empty((33,), dtype=torch.bool))  # 33 bytes
+    descr_d = TensorDescriptor.from_tensor(torch.empty((0,), dtype=torch.int64))  # 0 bytes
+    descr_e = TensorDescriptor.from_tensor(torch.empty((96, 8), dtype=torch.bfloat16))  # 1536 bytes
+    descr_f = TensorDescriptor.from_tensor(torch.empty((1792,), dtype=torch.uint8))  # 1792 bytes
+
+    # TODO test:
+    # - max_alloc_timeout in __init__
+    # - ensure that alloc_timeout 0 never waits
+    # - check that canceling before alloc does not trigger that alloc eventually
+    # - ensure that alloc_timeout 0 always allocates if free memory
```
justheuristic, line 95:
```suggestion
```

src/petals/server/memory_cache.py
```diff
@@ -60,11 +61,14 @@ def handle_counter(self, value: int):
         self._handle_counter.value = value
 
     @contextlib.asynccontextmanager
-    async def allocate_cache(self, *descriptors: TensorDescriptor) -> AsyncContextManager[Sequence[Handle]]:
+    async def allocate_cache(
+        self, *descriptors: TensorDescriptor, timeout: Optional[float] = None
```
borzunov, line 65:
```suggestion
        self, *descriptors: TensorDescriptor, timeout: float
```

Let's assume `timeout` to be known here, otherwise we'll have defaults in two places.

src/petals/server/memory_cache.py
```diff
@@ -74,6 +78,8 @@ async def allocate_cache(self, *descriptors: TensorDescriptor) -> AsyncContextMa
         """
         assert os.getpid() != self.runtime_pid, "must be called by a ConnectionHandler, not runtime"
         assert all(descr.device is not None for descr in descriptors), "please specify allocated devices"
+        if self.max_alloc_timeout is not None:
+            timeout = min(timeout, self.max_alloc_timeout) if timeout is not None else self.max_alloc_timeout
```
borzunov, line 82:
```suggestion
            timeout = min(timeout, self.max_alloc_timeout)
```

Let's assume timeout to be known here, otherwise we'll have defaults in two places. Notably, the default here was different from the zero default in `backend.py`, which was confusing.

src/petals/server/memory_cache.py
```diff
@@ -84,41 +90,67 @@ async def allocate_cache(self, *descriptors: TensorDescriptor) -> AsyncContextMa
             f"already used {cur_size / gib:.2f}/{friendly_max_size} GiB ({cur_size / max_size * 100:.1f}%)"
         )
 
-        alloc_task = asyncio.create_task(self._schedule_alloc(max_alloc_size, *descriptors))
+        alloc_task = asyncio.create_task(self._schedule_alloc(max_alloc_size, *descriptors, timeout=timeout))
         try:
             handles = await shield_and_wait(alloc_task)
-            logger.info(f"rpc_inference.alloc(size={max_alloc_size / gib:.2f} GiB)")
+            logger.info(f"rpc_inference.alloc-done(size={max_alloc_size / gib:.2f} GiB)")
```
borzunov, line 96:
```suggestion
            logger.info(f"rpc_inference.alloc_done(size={max_alloc_size / gib:.2f} GiB)")
```

Please stick to the style of all similar log messages

src/petals/server/memory_cache.py
```diff
@@ -84,41 +90,67 @@ async def allocate_cache(self, *descriptors: TensorDescriptor) -> AsyncContextMa
...
             yield handles
         finally:
-            self._free(max_alloc_size, alloc_task)
+            await shield_and_wait(self._schedule_free(max_alloc_size, alloc_task))
```
borzunov, line 99: **Important:** You can't do `await` in finally, this leads to deadlocks that we just fixed in #396.

Freeing memory should always happen instantly - why do you need that?

src/petals/server/memory_cache.py
```diff
@@ -84,41 +90,67 @@ async def allocate_cache(self, *descriptors: TensorDescriptor) -> AsyncContextMa
...
+                finally:
+                    self._cache_overfull_event.clear()
+            yield
+
+    async def _schedule_free(self, alloc_size: int, alloc_task: asyncio.Task):
+        """
+        This method should be called inside asyncio.shield() because:
+            - hivemind.utils.enter_asynchronously() does not always release the lock on cancellation
+            - _schedule_free() must finish freeing memory even in case of cancellation
+        """
```
borzunov, line 153: Please revert this change - this function cannot `await` and **must** return instantly (see comments above).

```suggestion
    def _free(self, alloc_size: int, alloc_task: asyncio.Task):
```

src/petals/utils/misc.py
```diff
@@ -5,3 +5,13 @@
 
 def is_dummy(tensor: torch.Tensor):
     return tensor.numel() == 0
+
+
+SPECIAL_DTYPE_SIZES = {torch.bool: 1, torch.int8: 1, torch.qint32: 4}
```
borzunov, line 10:
```suggestion
SPECIAL_DTYPE_SIZES = {torch.bool: 1}
```

`int8` and `qint32` are supported by `iinfo`.

src/petals/server/memory_cache.py
```diff
@@ -84,41 +90,67 @@ async def allocate_cache(self, *descriptors: TensorDescriptor) -> AsyncContextMa
...
+        except TimeoutError:
+            raise AllocationFailed(f"Could not allocate {alloc_size} (timeout={timeout})")
 
+    @contextlib.asynccontextmanager
+    async def _wait_for_free_memory(self, alloc_size: int, timeout: Optional[float]):
+        start_time = time.perf_counter()
         loop = asyncio.get_event_loop()
-        async with hivemind.utils.enter_asynchronously(self._lock_acquire_memory):
+        if timeout == 0:  # if waiting is not allowed, fail when you or anyone else begins waiting
+            stop_when_completes = loop.run_in_executor(None, self._cache_overfull_event.wait)
```
borzunov, line 133: **Important:** If waiting is not allowed, please either allocate or return exception right away. Please write code explicitly doing that - otherwise it's very difficult to check this case and we may end up with deadlocks.

src/petals/server/memory_cache.py
```diff
@@ -172,5 +204,32 @@ def use_cache(self, *handles: Handle) -> Sequence[torch.Tensor]:
...
+    aenter_task = asyncio.create_task(anext(aenter_and_aexit))
+    stop_task = asyncio.create_task(_wait_for_deadline())
+    try:
+        await asyncio.wait({aenter_task, stop_task}, return_when=asyncio.FIRST_COMPLETED)
+        if stop_task.done():
+            raise TimeoutError("Did not enter context in time")
+        yield aenter_task.result()
+    finally:
+        if aenter_task.done():
+            await anext(aenter_and_aexit)  # exit normally
```
borzunov, line 229: **Important:** `await` in finally does not always work (see above comments and #396). Can we simplify this function without it?

src/petals/server/memory_cache.py
```diff
@@ -84,41 +90,67 @@ async def allocate_cache(self, *descriptors: TensorDescriptor) -> AsyncContextMa
...
+    async def _wait_for_free_memory(self, alloc_size: int, timeout: Optional[float]):
+        start_time = time.perf_counter()
         loop = asyncio.get_event_loop()
-        async with hivemind.utils.enter_asynchronously(self._lock_acquire_memory):
+        if timeout == 0:  # if waiting is not allowed, fail when you or anyone else begins waiting
+            stop_when_completes = loop.run_in_executor(None, self._cache_overfull_event.wait)
+        else:  # otherwise, only fail if timeout expires
+            stop_when_completes = asyncio.sleep(timeout if timeout is not None else float("inf"))
+
+        async with wait_for_aenter(enter_asynchronously(self._lock_acquire_memory), stop_when_completes):
```
borzunov, line 137: **Important:** You don't need `wait_for_aenter()` here (which is super complicated and is likely to have bugs with cancellations, see below). You can instead do:

```
from async_timeout import timeout  # This module is already used in Petals

async with timeout(timeout):  # timeout == None is correctly supported here
    async with enter_asynchronously(self._lock_acquire_memory):
        # do stuff
```

src/petals/server/memory_cache.py
```diff
@@ -172,5 +204,32 @@ def use_cache(self, *handles: Handle) -> Sequence[torch.Tensor]:
         yield tuple(self._allocated_tensors[handle] for handle in handles)
 
 
+@contextlib.asynccontextmanager
+async def wait_for_aenter(context: contextlib.AbstractAsyncContextManager, stop_when_completes: Coroutine):
```
borzunov, line 208: This function is not needed since the code using it may be rewritten in a simpler way (see above).

borzunov (changes requested):
Thanks for the PR!

The current code is complicated and have places that are likely to introduce bugs. Please follow the comments on how to simplify the code and address potential issues.

src/petals/server/memory_cache.py
justheuristic: Here, timeout=None is not a default to be replaced, it literally means no timeout

src/petals/server/memory_cache.py
justheuristic: It looks like we have a misunderstanding: this is not a default. The suggested code would not support max_alloc_timeout=None

src/petals/server/memory_cache.py
justheuristic: it's gone

src/petals/server/memory_cache.py
justheuristic: implemented this suggestion with minor alterations to ensure fairness

src/petals/server/memory_cache.py
justheuristic: it's gone, good catch!

src/petals/server/memory_cache.py
justheuristic: it's gone, replaced with a double-checked lock now

src/petals/utils/misc.py
justheuristic: Alas, it is not:
image

src/petals/server/handler.py
```diff
@@ -528,14 +531,18 @@ def _check_uids(self, uids: str) -> Tuple[ModuleUID, ...]:
 
     @contextlib.asynccontextmanager
     async def _allocate_cache(
-        self, backends: Sequence[TransformerBackend], batch_size: int, max_length: int
+        self,
+        backends: Sequence[TransformerBackend],
+        batch_size: int,
```
borzunov, line 536: nit:

```suggestion
        *,
        batch_size: int,
```

So nobody passes `max_length` instead of `batch_size`, etc.

src/petals/server/handler.py
```diff
@@ -166,7 +167,9 @@ async def rpc_inference(
 
                 batch_size = request.tensors[0].size[0] if request.tensors else 1
 
-                async with self._allocate_cache(requested_backends, batch_size, max_length) as cache_handles:
+                async with self._allocate_cache(
+                    requested_backends, batch_size, max_length, alloc_timeout
```
borzunov, line 171: nit:

```suggestion
                    requested_backends, batch_size=batch_size, max_length=max_length, timeout=alloc_timeout
```

src/petals/server/memory_cache.py
borzunov: Can we still drop `None` support here and require passing `float('inf')` for this case? The metadata's timeout is casted to float in the wrapping code anyway

src/petals/server/memory_cache.py
```diff
@@ -96,28 +111,58 @@ def get_allocation_size(*descriptors: TensorDescriptor) -> int:
...
-                self._pipe_send.send((handles, descriptors))
-                return handles
-
-    def _free(self, alloc_size: int, alloc_task: asyncio.Task) -> None:
+
+        self.enqueued_size_bytes += alloc_size
+        allocated = False
+        try:
+            async with async_timeout.timeout(timeout if timeout != 0 else self.max_alloc_timeout):
+                # the code below uses double-checked locking: https://en.wikipedia.org/wiki/Double-checked_locking
```
borzunov, line 145:
```suggestion
```

Apparently, this is not used anymore

src/petals/server/memory_cache.py
```diff
@@ -96,28 +111,58 @@ def get_allocation_size(*descriptors: TensorDescriptor) -> int:
...
-                self.handle_counter += len(handles)  # note: this will eventually overflow and it is okay
-                self._pipe_send.send((handles, descriptors))
-                return handles
-
-    def _free(self, alloc_size: int, alloc_task: asyncio.Task) -> None:
+
+        self.enqueued_size_bytes += alloc_size
+        allocated = False
+        try:
+            async with async_timeout.timeout(timeout if timeout != 0 else self.max_alloc_timeout):
```
borzunov, line 144:
```suggestion
            context_manager = async_timeout.timeout(timeout) if timeout else contextlib.AsyncExitStack()
            # contextlib.AsyncExitStack() is used as a null context here
            async with context_manager:
```

src/petals/server/memory_cache.py
```diff
@@ -96,28 +111,58 @@ def get_allocation_size(*descriptors: TensorDescriptor) -> int:
...
+                            raise AllocationFailed(f"Could not allocate {alloc_size} bytes immediately: out of memory")
+                        elapsed_time = time.perf_counter() - start_time
+                        remaining_timeout = max(0.0, timeout - elapsed_time) if timeout is not None else None
+                        await loop.run_in_executor(None, self._wait_until_available, alloc_size, remaining_timeout)
+
+                allocated = True
+                self.enqueued_size_bytes -= alloc_size
+                yield
+        except (TimeoutError, asyncio.TimeoutError) as e:
+            raise AllocationFailed(f"Could not allocate {alloc_size} within {timeout} seconds ({e})")
```
borzunov, line 160: nit:

```suggestion
        except asyncio.TimeoutError:
            raise AllocationFailed(f"Could not allocate {alloc_size} within {timeout} seconds")
```

tests/test_cache.py
```diff
@@ -0,0 +1,166 @@
...
+        pass
+
+    async with cache.allocate_cache(_make_tensor_descriptor(100), timeout=999):
+        async with cache.allocate_cache(_make_tensor_descriptor(512), timeout=0):
+            async with cache.allocate_cache(_make_tensor_descriptor(128), _make_tensor_descriptor(32), timeout=1):
+                t_start = time.perf_counter()
+                with pytest.raises(AllocationFailed):
+                    async with cache.allocate_cache(_make_tensor_descriptor(768), timeout=0.1):
+                        pass
+                assert 0.1 < time.perf_counter() - t_start < 0.2, "wait time exceeds alloc timeout"
```
borzunov, line 38: Why isn't it `0.1 + eps`?

tests/test_cache.py
```diff
@@ -0,0 +1,166 @@
...
+                async with cache.allocate_cache(_make_tensor_descriptor(128), timeout=None):
+                    pass
+
+                t_start = time.perf_counter()
+                with pytest.raises(AllocationFailed):
+                    async with cache.allocate_cache(_make_tensor_descriptor(384), timeout=1.0):  # exceeds max timeout
+                        pass
+                assert 0.5 < time.perf_counter() - t_start < 0.6, "wait time exceeds max alloc timeout"
```
borzunov, line 46: Why 0.5, not 1.0?

src/petals/server/handler.py
justheuristic: +, will do

src/petals/server/handler.py
justheuristic: +, will do

src/petals/server/memory_cache.py
justheuristic: will do

tests/test_cache.py
justheuristic, line 46: (explained)

src/petals/server/memory_cache.py
justheuristic: i solemnly swear to test that it works with float('inf')

src/petals/server/memory_cache.py
```diff
@@ -96,28 +111,59 @@ def get_allocation_size(*descriptors: TensorDescriptor) -> int:
...
+    async def _schedule_alloc(
+        self, alloc_size: int, *descriptors: TensorDescriptor, timeout: float
+    ) -> Sequence[Handle]:
         """
         This method should be called inside asyncio.shield() because:
             - hivemind.utils.enter_asynchronously() does not always release the lock on cancellation
         """
+        try:
+            async with self._wait_for_free_memory(alloc_size, timeout):
+                async with enter_asynchronously(self._lock_metadata):
```
borzunov, line 127:
```suggestion
                with self._lock_metadata:
```

Before, we discussed that this lock is very short-living and there's no need to wait it in `async`

src/petals/server/memory_cache.py
```diff
@@ -12,24 +12,26 @@
 import time
 from typing import AsyncContextManager, Dict, Optional, Sequence
 
-import hivemind
+import async_timeout
 import torch
-from hivemind.utils import TensorDescriptor, get_logger
+from hivemind.utils import TensorDescriptor, enter_asynchronously, get_logger
```
borzunov, line 17:
```suggestion
from hivemind.utils import TensorDescriptor, get_logger
```

No need for this (see below)

borzunov (approved):
Please accept my last comment and merge once you're ready.

PR #435, opened 2023-08-04
[Refactor] extract block forward, backward and inference into a separate file

justheuristic:
This PR does not change any functionality. It merely moves stuff around.

__Problem:__ handler is huge and difficult to navigate. It is about to become more complicated once we support general peft

__Suggesion:__ move all math (forward and backward pass; grad w.r.t. prompt, etc) to a separate file. Handler handles all networking and queues, but not the math; block_methods is responsible for math only.

List of changes:
- [x] handler.py/_rpc_forward became block_methods/rpc_forward
- [x] handler.py/_rpc_backward became block_methods/rpc_backward
- [x] the math bits of rpc_inference were extracted into block_methods/iterate_rpc_inference

I took care to avoid any other changes


borzunov (approved):
Thanks for the PR.

Please rename it to `block_functions.py` and test that pushes are successfully sent (this is not covered by test and can break due to `can_push` refactor). Then it's good to go!

PR #438, opened 2023-08-06
Update install instructions to cuda 11.8

vadi2:
Update install instructions to cuda 11.8 - seems safe enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Petals is a community-run system &mdash; we rely on people sharing their GPUs. Y
 🐍 **Linux + Anaconda.** Run these commands:
 
 ```bash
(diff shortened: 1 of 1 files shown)


borzunov:
Hi @vadi2,

Thanks for the PR!

Could you please clarify why it's useful and if this drops support for any old GPUs?

vadi2:
I'm not sure if it drops support for any older GPUs, but it is useful because it gives a 2x speedup on Ada architectures: https://gpux.ai/blog/sdxl_50_faster

borzunov:
@vadi2 Got it! I'll double-check that with colleagues and get back to merging this.

PR #445, opened 2023-08-07
Add customizable input tensors

artek0chumak:
This PR includes a new way to provide tensors to the servers. Instead of a strict list of tensors described on the server side, clients can send any tensors to the servers.


src/petals/server/handler.py
```diff
@@ -430,6 +436,7 @@ async def rpc_backward(self, request: runtime_pb2.ExpertRequest, context: P2PCon
             metadata = MSGPackSerializer.loads(request.metadata) if request.metadata else {}
             active_adapter = self._get_active_adapter(metadata)
             points = metadata.get("points", 0)
+            tensor_names = metadata.get("tensor_names", None)
```
mryab, line 439: Just in case, None is the default return value of `get`

src/petals/client/routing/sequence_manager.py
```diff
@@ -461,10 +461,23 @@ def get_request_metadata(self, protocol: str, *args, **kwargs) -> Optional[Dict[
...
+        return dict(
+            points=self.policy.get_points(protocol, *args, **kwargs),
+            active_adapter=self.config.active_adapter,
+            tensor_names=self.get_tensor_names(protocol),
+        )
 
     def shutdown(self):
         self._thread.shutdown()
+        
+    def get_tensor_names(self, protocol):
```
justheuristic, line 473: I believe we were supposed to have those values in rpc_info; However, the values from rpc_info are de facto obsolete and it would take time to fix them.

Can you please add a todo to ensure that these values are consistent wit rpc_info?

src/petals/client/remote_forward_backward.py
```diff
@@ -84,26 +85,22 @@ async def run_remote_forward(
...
     inputs = tuple(tensor.cpu().detach() for tensor in forward_inputs)
+    assert len(inputs) >= len(args_schema) + 1, "Inputs and prompt tensors are necesseary for a forward step"
 
     # Asynchronous serialization
     loop = asyncio.get_running_loop()
     serialized_tensors = await asyncio.gather(
         *(
             loop.run_in_executor(None, serialize_torch_tensor, tensor.to(proto.dtype), proto.compression)
-            for tensor, proto in zip(inputs, nested_flatten(forward_schema_with_prompts))
+            for tensor, proto in zip(inputs, nested_flatten(forward_schema))
```
justheuristic, line 103: Do you really need to nested_flatten forward_schema? It looks flat to me

src/petals/server/block_functions.py
```diff
@@ -34,7 +36,10 @@ async def run_rpc_forward(
     :param requested_backends: a sequence of transformer blocks in the same order as they appear in forward pass
     :returns: hidden states after the last layer [batch_size, seq_length, hid_size]
     """
-    hidden_states, prompts = flat_tensors
+    if structure is not None:
+        flat_tensors, kwargs = unpack_args_kwargs(flat_tensors, structure)
```
justheuristic, line 40: It seems that kwargs is unused; If that is intended, please leave a TODO

src/petals/server/block_functions.py
```diff
@@ -72,8 +77,12 @@ async def run_rpc_backward(
     active_adapter: str = "",
     prioritizer: TaskPrioritizerBase,
     points: int = 0,
+    structure: Optional[Dict[str, Any]] = None,
 ) -> Union[torch.Tensor, Sequence[torch.Tensor]]:
-    inputs, grad_outputs, prompts = flat_tensors
+    if structure is not None:
+        flat_tensors, kwargs = unpack_args_kwargs(flat_tensors, structure)
```
justheuristic, line 83: If kwargs are not used intentionally (for this PR), please leave a TODO to actually forward them

src/petals/client/inference_session.py
```diff
@@ -18,11 +18,14 @@
...
 from hivemind.p2p import P2P
 from hivemind.proto import runtime_pb2
+from hivemind.utils.tensor_descr import BatchTensorDescriptor
 
 from petals.client.routing.sequence_manager import RemoteSequenceManager, SequenceManagerConfig, maybe_log_traceback
 from petals.data_structures import CHAIN_DELIMITER, ModuleUID, RemoteSpanInfo, RPCInfo
 from petals.server.handler import TransformerConnectionHandler
+from petals.utils.packaging import pack_args_kwargs
 from petals.utils.misc import DUMMY, is_dummy
+from petals.utils.packaging import pack_args_kwargs
```
mryab, line 28: The import appears to be duplicated

src/petals/client/inference_session.py
```diff
@@ -237,7 +251,10 @@ def _enter_server_sessions(self, chosen_spans: List[RemoteSpanInfo]) -> List[_Se
         try:
             for span in chosen_spans:
                 span_uids = CHAIN_DELIMITER.join(self._sequence_manager.block_uids[span.start : span.end])
-                metadata = self._sequence_manager.get_request_metadata("rpc_inference", span_uids, peer_id=span.peer_id)
+                structure = dict()
+                metadata = self._sequence_manager.get_request_metadata(
+                    "rpc_inference", span_uids, structure, peer_id=span.peer_id
+                )
```
mryab, line 257: Is it correct that `structure` does not get populated? Is this something we need to do in the future or is passing an empty dict here intentional?

src/petals/client/remote_forward_backward.py
```diff
@@ -84,26 +86,22 @@ async def run_remote_forward(
...
-        raise TypeError(f"Inputs do not match expert input schema. Did you pass the right number of parameters?")
-
-    forward_inputs = nested_flatten(forward_inputs)
+    compression = args_schema[0].compression
+    forward_schema = tuple(
+        BatchTensorDescriptor.from_tensor(arg, compression)
+        for arg in forward_inputs
+    )
     inputs = tuple(tensor.cpu().detach() for tensor in forward_inputs)
+    assert len(inputs) >= len(args_schema) + 1, "Inputs and prompt tensors are necessary for a forward step"
```
mryab, line 97:
```suggestion
    assert len(inputs) > len(args_schema), "Inputs and prompt tensors are necessary for a forward step"
```

mryab: If we assert that we need a specific number of arguments (and specific kind of arguments), maybe we can check the inputs more explicitly?

src/petals/utils/packaging.py
```diff
@@ -0,0 +1,42 @@
...
+def get_tensor_index(item: bytes):
+    return int(item[3:])
+
+
+def pack_args_kwargs(*args, **kwargs):
+    masked_flat_values, flat_tensors, tensor_to_index = [], [], {}
+    for value in nested_flatten((args, kwargs)):
+        if isinstance(value, torch.Tensor):
+            tensor_index = tensor_to_index.setdefault(value, len(flat_tensors))
+            if tensor_index == len(flat_tensors):
```
mryab, line 27: Not sure if I understand: is this necessary to deduplicate input tensors (e.g. DUMMY)?

src/petals/client/remote_forward_backward.py
artek0chumak: Maybe in the future. Currently, I want to have a better way to provide tensors from clients into servers. I don't want to enhance args_scheme in this PR, but it will be awesome todo.

I will leave a TODO about it here.

borzunov: @mryab I think it's more clear the way @artek0chumak wrote it - so that no one misses the `>` instead of `>=`, and everyone sees that the min number of args is `len(args_schema) + 1`.

src/petals/utils/packaging.py
```diff
@@ -0,0 +1,50 @@
...
+
+# TODO: Move functions to hivemind
+
+
+def _mark_masked_tensor(index: int):
+    return b"__T" + str(index).encode()
+
+
+def _is_masked_tensor(item: Any):
+    return isinstance(item, bytes) and re.match(b"^__T\d+$", item) is not None
```
borzunov, line 15:
```suggestion
    return isinstance(item, bytes) and item.startswith("__T") is not None
```

I strongly recommend to avoid `re` to increase performance. If you keep it, please use `re.fullmatch(rb"__T\d+", item)` instead (does the same thing but won't fail if a user adds new `\smth` sequence).

src/petals/utils/packaging.py
```diff
@@ -0,0 +1,50 @@
+import re
+from typing import Any
+
+import torch
+from hivemind import nested_flatten, nested_pack
+
+# TODO: Move functions to hivemind
+
+
+def _mark_masked_tensor(index: int):
```
borzunov, line 10:
```suggestion
def _mark_masked_tensor(index: int) -> bytes:
```

Please complete the type annotations at least in the cases where the type is known for sure.

src/petals/utils/packaging.py
borzunov: Yes, @justheuristic wanted to dedup large tensors

tests/test_aux_functions.py
```diff
@@ -44,3 +47,32 @@ def test_compute_throughput(inference: bool, n_tokens: int, tensor_parallel: boo
...
+
+    restored_args, restored_kwargs = unpack_args_kwargs(flat_tensors, tensor_structure)
+
+    assert len(restored_args) == len(args)
+    assert torch.all(restored_args[0] == x).item() and restored_args[2] is None
+    assert nested_compare((args, kwargs), (restored_args, restored_kwargs))
+    for original, restored in zip(nested_flatten((args, kwargs)), nested_flatten((restored_args, restored_kwargs))):
+        if isinstance(original, torch.Tensor):
+            if is_dummy(original):
+                assert is_dummy(restored)
```
borzunov, line 74: Why won't `torch.all()` work for the dummies? If it does, please remove this edge case

tests/test_aux_functions.py
```diff
@@ -44,3 +47,32 @@ def test_compute_throughput(inference: bool, n_tokens: int, tensor_parallel: boo
...
+            else:
+                assert torch.all(original == restored)
```
borzunov, line 76:
```suggestion
            assert torch.all(original == restored)
```

`torch.all()` should work for dummies

src/petals/client/inference_session.py
```diff
@@ -144,13 +139,24 @@ def step(
...
                 request_metadata["next_servers"] = next_servers
 
+        request_metadata["tensor_structure"] = tensor_structure
+
+        server_side_inference_schema, kwargs_schema = self.rpc_info["inference_schema"]
+        compression = server_side_inference_schema[0].compression
+        inference_schema = tuple(BatchTensorDescriptor.from_tensor(arg, compression) for arg in input_tensors)
+
+        # TODO: create more explicit way to check servers schema and client's structure
+        assert len(input_tensors) + (is_dummy(hypo_ids) and is_dummy(prompts)) >= len(
```
borzunov, line 149:
```suggestion
        assert len(input_tensors) + >= len(
```

Why add dummies separately here? Aren't they a part of `input_tensors` already?

src/petals/client/inference_session.py
```diff
@@ -144,13 +139,24 @@ def step(
             if next_servers:
                 request_metadata["next_servers"] = next_servers
 
+        request_metadata["tensor_structure"] = tensor_structure
+
+        server_side_inference_schema, kwargs_schema = self.rpc_info["inference_schema"]
+        compression = server_side_inference_schema[0].compression
+        inference_schema = tuple(BatchTensorDescriptor.from_tensor(arg, compression) for arg in input_tensors)
```
borzunov, line 146: Can we use `nested_flatten()` here as the previous code does? This way, we'll support different compressions for each tensor and the code should become simpler

borzunov (approved):
Nice work, thanks for the PR! I've left a few minor comments, please merge when you:

- [x] Address all the comments
- [x] Test compatibility between old clients and new servers
- [x] Test compatibility between new clients and old servers (this can be tested using a local client and the existing servers in the public swarm without any risks)

artek0chumak:
Check backward capability:
- Old client -- new server (colab + private swarm with bloom-560m on a server). Forward, backward and inference sessions were running without issues.
- New client -- old server (public swarm, stabilityai/StableBeluga2). Forward, backward and inference sessions were running without issues.

src/petals/utils/packaging.py
```diff
@@ -0,0 +1,49 @@
...
+
+# TODO: Move functions to hivemind
+
+
+def _mark_masked_tensor(index: int) -> bytes:
+    return b"__T" + str(index).encode()
+
+
+def _is_masked_tensor(item: Any) -> bool:
+    return isinstance(item, bytes) and item.startswith(b"__T") is not None
```
borzunov, line 14:
```suggestion
    return isinstance(item, bytes) and item.startswith(b"__T")
```

Bug: it does return False/True

borzunov (approved):

src/petals/client/inference_session.py
borzunov: Ok

PR #452, opened 2023-08-08
Show argparse defaults, test petals.cli.run_dht

borzunov:


src/petals/server/server.py
```diff
@@ -78,7 +78,7 @@ def __init__(
         sender_threads: int = 1,
         balance_quality: float = 0.75,
         mean_balance_check_period: float = 120,
-        mean_block_selection_delay: float = 2.5,
+        mean_block_selection_delay: float = 5,
```
borzunov, line 81: Previous delay was not enough - servers were often choosing the same blocks since they didn't have time to write DHT "JOINING" messages.

PR #461, opened 2023-08-14
Support repetition_penalty in generation

borzunov:


borzunov:
Deprecated in favor of #464.

PR #463, opened 2023-08-14
Move SequenceManagerConfig -> ClientConfig, petals.dht_utils -> petals.utils.dht

borzunov:
This PR moves `SequenceManagerConfig` to a separate file and renames it to `ClientConfig` - since it contains many args not specific to `SequenceManager`.


justheuristic (approved):
Everything looks fine, thanks!
Since someone can conceivably use this class, I'd recommend leaving a temporary alias with a deprecation warning  (SequenceManagerConfig = ClientConfig)

borzunov:
Good idea!

Issue #460, opened 2023-08-14
Add repetition_penalty to RemoteGenerationMixin


Webifi:
RemoteGenerationMixin / src/petals/client/remote_generation.py doesn't appear to have support for repetition_penalty.  

Is it possible have this added?

borzunov:
Hi @Webifi, I had a branch implementing that, I'll look it up today

borzunov:
Hi @Webifi, I have some old code implementing it: https://github.com/bigscience-workshop/petals/pull/461

I'll update it to be compatible with the latest code and let you know!

borzunov:
The current implementation is not compatible with `top_k` and `top_p` though. We'd like to fix that by introducing a broader compatibility with built-in HF Transformers generation algorithms at some point.

borzunov:
@Webifi I've actually decided to implement a broad compatibility with all the HF `.generate()` options instead (including `repetition_penalty`). Working on it in #464. I'll let you know when it's done and merged.

borzunov:
@Webifi It's done. I've also added support to chat.petals.dev in https://github.com/petals-infra/chat.petals.dev/commit/9e4e06b982e24f190c9fa9fe5f9947b6fc1a98b4.

You can try it out by opening https://chat.petals.dev and trying different values in the JS dev console:

```js
generationParams.repetition_penalty = 0.1; // Always repeat

generationParams.repetition_penalty = 1.2; // Recommended value in the repetition_penalty paper

generationParams.repetition_penalty = 10.0; // Never repeat
```

PR #464, opened 2023-08-14
Make client compatible with transformers' GenerationMixin

borzunov:
This PR drops custom generation codes and introduces compatibility with `transformers.GenerationMixin` instead. This includes support for more sampling options (`top_p`, `top_k`, `repetition_penalty` requested in #460) and beam search - all that is now identical to running model with transformers locally.

Most features (excluding beam search and other rarely used stuff) are also compatible with resuming existing sessions.

### Breaking changes

If `.generate()` or forward passes are being run inside an `.inference_session()` context, they now use the opened session by default. So, these snippets are now equivalent:

```python
# Using default session
with model.inference_session(max_length=100):
    output_ids = model.generate(input_ids, max_new_tokens=3)

# Explicitly specifying a session
with model.inference_session(max_length=100) as sess:
    output_ids = model.generate(input_ids, max_new_tokens=3, session=sess)
```

Earlier, the 1st snippet was creating a new session, which is not what most people expected (= such code was most likely to introduce a bug, which is now fixed).


src/petals/models/bloom/model.py
```diff
@@ -59,21 +58,34 @@ def forward(
         else:
             raise ValueError("You have to specify either input_ids or inputs_embeds")
 
+        # The causal mask will be added on the server-side
+        assert (
+            attention_mask is None or (attention_mask == 1).all()
+        ), f"Custom attention masks are not supported, {attention_mask=}"
+        assert head_mask is None, f"Custom head masks are not supported, {head_mask=}"
+        assert use_cache is None or use_cache, f"{use_cache=} is not supported"
```
artek0chumak, line 66: Is it better to get an assertion by use_cache? I prefer warning or checking of inference sessions.

src/petals/client/remote_generation.py
```diff
@@ -1,349 +1,105 @@
...
+    @docstring_from(RemoteSequential.use_session)
+    def use_session(self, session: Optional[InferenceSession]) -> ContextManager[InferenceSession]:
+        return self.transformer.h.use_session(session)
 
+    @docstring_from(RemoteSequential.inference_session)
+    def inference_session(self, **kwargs) -> ContextManager[InferenceSession]:
         return self.transformer.h.inference_session(**kwargs)
 
-    @torch.inference_mode()
+    @docstring_from(transformers.GenerationMixin.generate.__doc__)
```
artek0chumak, line 54: Add checking for available generation algorithms.

borzunov: I've added a warning that beam search won't give the same results on a resumed inference session. I don't want to check everything right now but will add known issues. Most importantly, all the previous functionality works right now.

tests/test_full_model.py
```diff
@@ -53,138 +63,94 @@ def test_full_model_exact_match(tokenizer, use_peft, pass_empty_tensors, atol_fo
...
-        model, input_ids=hf_inputs, max_length=inputs.size(1) + max_new_tokens, beam_scorer=beam_scorer
-    )
-    assert torch.allclose(remote_outputs, hf_outputs), "Beam search results are not identical to HF"
+def test_beam_search_generation(tokenizer, model, ref_model, max_new_tokens=4, num_beams=5):
+    inputs = tokenizer("A cat sat on a mat", return_tensors="pt")["input_ids"]
+
+    options = dict(max_new_tokens=max_new_tokens, num_beams=num_beams, do_sample=False)
+    outputs = make_generate_calls(model, inputs, **options)
+    ref_outputs = ref_model.generate(inputs, **options)
+    assert torch.allclose(outputs, ref_outputs), f"Beam search results are not identical to HF with {multiple_calls=}"
```
justheuristic, line 156: multple_calls is not defined. Please set up a parameter (multiple_calls) via pytest.parameterize to test that we 
(at least) do not raise an error here

src/petals/models/bloom/model.py
```diff
@@ -39,16 +39,15 @@ def __init__(self, config: DistributedBloomConfig, *, dht: Optional[hivemind.DHT
     def forward(
         self,
         input_ids: Optional[torch.LongTensor] = None,
-        inputs_embeds: Optional[torch.Tensor] = None,
+        past_key_values: Optional[RemotePastKeyValues] = None,
```
justheuristic, line 42: Not a request, but a sanity check: when we support prefix_tuning, this logic will need to change. Superficially, it should be simple to support both RemotePastKeyValues and nested pytorch tensors as past_key_values.
__Can you think of some caveats?__

For instance, how do we pass `past_key_values` tensors from prefix_tuning during .generate , when inferencing a prefix-tuned model?

justheuristic (approved):

src/petals/models/bloom/model.py
borzunov: Prefix-tuning past kvs are only used on the 1st generation step, while mocked `RemotePastKeyValues` here are only used for 2nd step and later to indicate that we have **some** past kvs (it's not important which ones for HF's `.generate()`). So I don't think these things contradict each other or could introduce any conflicts.

tests/test_full_model.py
borzunov: I've just fixed the undefined var for now. I've manually checked that the warning is shown but don't want to commit to this behavior in tests.

PR #467, opened 2023-08-17
Forward arbitrary kwargs

justheuristic:
NB: this pull request makes several drastic changes to the backend, block_functions and pools. If it interferes with long-term plans for the codebase, please raise a concern - i'm happy to rollback any detrimetnal changes.

Feature
- [x] servers will now support arbitrary kwargs for transformers blocks
- [ ] servers can now backprop w.r.t. additional keyword args and return gradients

Internal codebase changes:
- [ ] TransformerBackend no longer inherits hivemind ModuleBackend
    - ... and therefore does not trigger (empty) optimizer & scheduler, saving/loading the (unused) training state, etc
    -  as of right now, 50% ModuleBackend methods are rewritten in TransformerBackend and the other 50% are unused

- [x] Task and PrioritizedTaskPool support kwargs
    - note: if we (eventually) implement server-side batching, we can only batch queries if they have the same input schema anyway
        , and therefore, this pull request does not make server-side batching any more complicated than it already is

Tests:
- [ ] everything that used to work works again
- [ ] forward with extra kwargs
- [ ] inference with attention mask **chunking**
- [ ] full model exact match with attention mask
- [ ] block-level backward with non-differentiable kwargs
- [ ] block-level backward with differentiable kwargs

### DIFF
--- a/src/petals/server/backend.py
+++ b/src/petals/server/backend.py
@@ -5,7 +5,7 @@ from itertools import chain
 from typing import Any, Dict, Optional, Sequence, Tuple, Union
 
 import torch
-from hivemind import BatchTensorDescriptor, TensorDescriptor
+from hivemind import BatchTensorDescriptor, TensorDescriptor, nested_flatten, nested_map
 from hivemind.moe.expert_uid import ExpertUID
 from hivemind.moe.server.module_backend import ModuleBackend
 from hivemind.utils import get_logger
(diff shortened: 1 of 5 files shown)


justheuristic:
note 2 self: __old client runs backward with inputs that do not require_grad, we must support that!__

justheuristic:
note 2self: after i wake up
- [ ] add args/kwargs partitioning in _RemoteSequentialAutogradFunction
- [ ] add args/kwargs forwarding in _ServerInferenceSession
- [ ] modify _ServerInferenceSession to use user-defined compression, metadata, schema, etc
- [ ] modify inference_session.py to save past args/kwargs and re-send them on server failure

src/petals/client/sequential_autograd.py
```diff
@@ -141,7 +145,7 @@ async def sequential_backward(
             try:
                 if attempt_no >= 1:
                     _, backup_inputs, backup_sequences = await sequential_forward(
-                        inputs, prompts, sequence_manager, start_index=span.start, end_index=span.end
+                        sequence_manager, inputs, prompts, start_index=span.start, end_index=span.end
```
justheuristic, line 148: subjective matter: sequence_manager is the first parameter to most internal functions; can rollback if the reviewer disagrees.

src/petals/server/backend.py
```diff
@@ -178,6 +186,17 @@ def _update_cache_inplace(
...
+    def _select_kwargs_chunk(kwargs: Dict[str, Any], seq_len: int, offset: int, max_chunk_length: int):
+        if offset == 0 and max_chunk_length >= seq_len:
+            return kwargs
+        kwargs_chunk = {}
+        for key, value in kwargs.items():
+            if isinstance(value, torch.Tensor) and value.ndim >= 2 and value.shape[-2] == seq_len:
+                value = value[:, offset : offset + max_chunk_length]
+            kwargs_chunk[key] = value
+        return kwargs_chunk
+
```
justheuristic, line 199: Note: this is a potential problem; not all tensors where shape[-2] == seq_len can be time-sliced.

Counter-example: a LoRA adapter might accidentally have it's rank equal to sequence length

src/petals/client/sequential_autograd.py
```diff
@@ -227,15 +222,17 @@ class _RemoteSequentialAutogradFunction(torch.autograd.Function):
     """
 
     @staticmethod
-    def forward(ctx, inputs: torch.Tensor, prompts: torch.Tensor, sequence_manager: RemoteSequenceManager):
+    def forward(ctx, sequence_manager: RemoteSequenceManager, inputs: torch.Tensor, prompts: torch.Tensor):
+        # TODO add kwargs here; figure out a way to split kwargs across servers
```
justheuristic, line 226: problem: how do we split args/kwargs into sub-batches?

dvmazur:
@justheuristic solemnly swears to
- show a proof that forwarding kwargs works in a **basic test that is easy to follow**
- show an example of how new clients can work with old servers, at least for all supported basic ops
- show an example of how old clients can work with new servers

PR #487, opened 2023-08-30
Fix race condition in MemoryCache

borzunov:
See https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Value


justheuristic (approved):
thanks for the fix

PR #496, opened 2023-09-02
Force use_cache=True

borzunov:
Petals supports only `use_cache=True` for inference.

However, we should not reject `use_cache=False` since it returns identical results (just forces the slower O(n^3) inference algorithm instead of the O(n^2) one).

I allow `use_cache=False` since some models use this setting for reasons unclear to me (see https://huggingface.co/garage-bAInd/Platypus2-70B-instruct/discussions/8), and this led to `AssertionError` before this PR.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ install_requires =
     transformers>=4.32.0,<5.0.0  # if you change this, please also change version assert in petals/__init__.py
     speedtest-cli==2.1.3
     pydantic>=1.10,<2.0  # 2.0 is incompatible with hivemind yet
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -40,7 +40,7 @@ install_requires =
     transformers>=4.32.0,<5.0.0  # if you change this, please also change version assert in petals/__init__.py
     speedtest-cli==2.1.3
     pydantic>=1.10,<2.0  # 2.0 is incompatible with hivemind yet
-    hivemind @ git+https://github.com/learning-at-home/hivemind
+    hivemind==1.1.10.post2
```
borzunov, line 43: They are currently identical.

PR #504, opened 2023-09-05
delete duplicate tests

justheuristic:
This PR removes test_chained_calls. It consisted of two tests that duplicate test_remote_sequential.


justheuristic:
Closed: no consensus between maintainers whether or not we should merge it

PR #517, opened 2023-09-19
Fix file locks in NFS-mounted directories

### LINKED ISSUE #515: Unable to acquire fcntl.LOCK_SH lock when cache directory is mounted on NFS
Error showing failed to acquire shared file lock during downloading model weight file from hub when cache directory is on a NFS mount.

```pythonSep 19 15:19:28.075 [WARN] [petals.server.from_pretrained._load_state_dict_from_file:151] Cache for file pytorch_model-00013-of-00015.bin is corrupted, it will be downloaded again
Traceback (most recent call last):
  File "/root/petals/src/petals/server/from_pretrained.py", line 139, in _load_state_dict_from_file
    with allow_cache_reads(cache_dir):
  File "/usr/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/root/petals/src/petals/utils/disk_cache.py", line 26, in _blocks_lock
    fcntl.flock(lock_fd.fileno(), mode)
OSError: [Errno 9] Bad file descriptor`
```

Reproduce:
Make sure open the file on a NFS mounted directory:
```python
import os
import fclnt
f = open("/nfs-mounted-dir/abc","wb")
fcntl.flock(f.fileno(), fcntl.LOCK_SH)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: [Errno 9] Bad file descriptor
``` 

Suggested resolution:
Open file in "wb+" mode

```python
import os
import fclnt
f = open("/nfs-mounted-dir/abc","wb+")
fcntl.flock(f.fileno(), fcntl.LOCK_SH)

```

tonywang16:
Fix #515 unable to acquire shared lock on NFS mounted directory

### DIFF
--- a/src/petals/server/throughput.py
+++ b/src/petals/server/throughput.py
@@ -56,7 +56,7 @@ def get_server_throughput(
 
     # We use the system-wide lock since only one process at a time can measure the host throughput
     os.makedirs(lock_path.parent, exist_ok=True)
-    with open(lock_path, "wb") as lock_fd:
+    with open(lock_path, "wb+") as lock_fd:
         logger.info("Loading throughput info")
         fcntl.flock(lock_fd.fileno(), fcntl.LOCK_EX)
         # The OS will release the lock when lock_fd is closed or the process is killed
--- a/src/petals/utils/disk_cache.py
+++ b/src/petals/utils/disk_cache.py
@@ -22,7 +22,7 @@ def _blocks_lock(cache_dir: Optional[str], mode: int):
     lock_path = Path(cache_dir, BLOCKS_LOCK_FILE)
 
     os.makedirs(lock_path.parent, exist_ok=True)
-    with open(lock_path, "wb") as lock_fd:
+    with open(lock_path, "wb+") as lock_fd:
         fcntl.flock(lock_fd.fileno(), mode)
         # The OS will release the lock when lock_fd is closed or the process is killed
         yield


borzunov:
Hi @tonywang16,

Thanks for this bug fix! I'm curious about how did you come to the `wb+` solution?

Also, can you please change the lock file mode to `wb+` in [throughput.py](https://github.com/bigscience-workshop/petals/blob/5ce4f1a1598b1fca9fe6bd30cfbd85aa99bce2c7/src/petals/server/throughput.py#L59) too?

tonywang16:
> Hi @tonywang16,
> 
> Thanks for this bug fix! I'm curious about how did you come to the `wb+` solution?
> 
> Also, can you please change the lock file mode to `wb+` in [throughput.py](https://github.com/bigscience-workshop/petals/blob/5ce4f1a1598b1fca9fe6bd30cfbd85aa99bce2c7/src/petals/server/throughput.py#L59) too?

Sure, updated the throughput.py.

The issue is due to it opens the file in "wb" (Open the file for writing) only mode on NFS mount.

Then it trying to acquire SHARED LOCK(LOCK_SH) which is commonly called a reader lock without read permission.  Adding "+" allows for both types of locks on the same file descriptor.

BTW, the exclusive lock fcntl.LOCK_EX having no issue on NFS mount.

PR #523, opened 2023-09-28
Fix retries during inference

borzunov:


src/petals/client/inference_session.py
```diff
@@ -359,6 +358,10 @@ def _update_sequence(self, server_idx: int, block_idx: int, attempt_no: int) ->
         # If there is a failed span, this code replaces it, otherwise it just adds new ones
         if server_idx < n_prev_spans:
             updated_sessions[0].history = self._server_sessions[server_idx].history
+            updated_sessions[0].position = self._position
```
borzunov, line 361: This is the fix

Issue #536, opened 2023-11-07
Upstream Changes makes the demo not work


hrQAQ:
I'm following the toturial `Windows + WSL` to start up a server on my own PC. uponrun python -m petals.cli.run_server petals-team/StableBeluga2

hrQAQ:
changelog from transformers Release 4.35.0

> ## Attention mask refactor
> We refactored the attention mask logic for major models in transformers. For instance, we removed padding_mask argument which was ambiguous for some users
> 
> Remove ambiguous padding_mask and instead use a 2D->4D Attn Mask Mapper by @patrickvonplaten in https://github.com/huggingface/transformers/pull/26792
> [Attention Mask] Refactor all encoder-decoder attention mask by @patrickvonplaten in https://github.com/huggingface/transformers/pull/27086

By running the following commands, my windows server is successfully started.

```bash
pip uninstall transformers
pip install transformers==4.34.0
```

So, I suggest modifying the [setup.cfg](https://github.com/bigscience-workshop/petals/blob/main/setup.cfg) to make minimal changes.

justheuristic:
Thank you for reporting this.
I'm fixing transformers version for now. In the meantime, we're working on fixing the problem in upstream

PR #545, opened 2023-11-21
replaced call to `_prepare_decoder_attention_mask()` with  `_prepare_4d_causal_attention_mask()`

poedator:
fixes issue #536


justheuristic:
Hi! I can confirm that `def test_input_ids`, specifically, passes in the main branch but fails here. This is most likely a bug. I can look into this *eventually*, but if you have time, please look into this yourself first.

PR #553, opened 2024-01-29
Add Mixtral models

artek0chumak:
Add [Mixtral models](https://huggingface.co/docs/transformers/model_doc/mixtral) as runnable models.

TODO
- [x] Wait for support of the new version of transformers  (>4.35)
  - [x] Add *normal* support for [Cache](https://github.com/huggingface/transformers/blob/main/src/transformers/cache_utils.py)
- [x] Check compatibility of larger models ([TinyMixtral](https://huggingface.co/SanjiWatsuki/TinyMixtral-32x248M), [Mixtral-8x7B](https://huggingface.co/mistralai/Mixtral-8x7B-v0.1)) on GPU
- [x] Refactor some code(see `# TODO`)
- [ ] Optimize code for improved speed (mb next PR?)

### DIFF
--- a/src/petals/models/mixtral/block.py
+++ b/src/petals/models/mixtral/block.py
@@ -12,10 +12,11 @@ from transformers.models.mixtral.modeling_mixtral import MixtralDecoderLayer, Mi
 
 class WrappedMixtralBlock(MixtralDecoderLayer):
     def __init__(self, config: MixtralConfig, layer_idx: int):
+        super().__init__(config, layer_idx)
+
         self._attn_implementation = config._attn_implementation
         self.sliding_window = config.sliding_window
         self.layer_idx = layer_idx
-        super().__init__(config, layer_idx)
 
     def forward(
         self,
--- a/src/petals/models/mixtral/block.py
+++ b/src/petals/models/mixtral/block.py
@@ -2,12 +2,12 @@ from typing import Optional, Tuple
 
 import torch
 from transformers import MixtralConfig
-from transformers.models.mixtral.modeling_mixtral import MixtralDecoderLayer, MixtralModel
+from transformers.cache_utils import DynamicCache
 from transformers.modeling_attn_mask_utils import (
     _prepare_4d_causal_attention_mask,
     _prepare_4d_causal_attention_mask_for_sdpa,
 )
-from transformers.cache_utils import DynamicCache
+from transformers.models.mixtral.modeling_mixtral import MixtralDecoderLayer, MixtralModel
 
 
 class WrappedMixtralBlock(MixtralDecoderLayer):
@@ -38,7 +38,9 @@ class WrappedMixtralBlock(MixtralDecoderLayer):
             _past_key_value = self._reorder_cache_from_bloom(past_key_value, batch_size, past_key_values_length)
             past_key_value = DynamicCache()
             for idx in range(self.layer_idx):
-                past_key_value.update(torch.empty(_past_key_value[0].size()), torch.empty(_past_key_value[1].size()), idx)
+                past_key_value.update(
+                    torch.empty(_past_key_value[0].size()), torch.empty(_past_key_value[1].size()), idx
+                )
             past_key_value.update(_past_key_value[0], _past_key_value[1], self.layer_idx)
 
         if self._attn_implementation == "flash_attention_2":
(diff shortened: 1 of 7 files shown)


src/petals/server/from_pretrained.py
```diff
@@ -51,7 +52,11 @@ def load_pretrained_block(
     torch_dtype = resolve_block_dtype(config, torch_dtype)
 
     with init_empty_weights():
-        block = config.block_class(config)
+        if config.block_class == WrappedMixtralBlock:
+            config = PreTrainedModel._autoset_attn_implementation(config)
```
artek0chumak, line 56: Called function: https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L1316

src/petals/server/backend.py
```diff
@@ -91,6 +91,8 @@ def get_inference_cache_descriptors(self, batch_size: int, max_length: int) -> S
         cache_tensors = []
         for device, num_heads in zip(self.module.devices, self.shard_num_heads):
             num_heads //= self.config.num_key_value_groups
+            if hasattr(self.config, "num_key_value_heads"):
+                num_heads = self.config.num_key_value_heads
```
mryab, line 95: Does this still work when the layer is sharded?

src/petals/models/mixtral/model.py
```diff
@@ -0,0 +1,177 @@
...
+    def word_embeddings_layernorm(self) -> nn.Module:  # For compatibility with RemoteGenerationMixin
+        return nn.Identity()
+
+    @property
+    def h(self) -> RemoteSequential:  # For compatibility with RemoteGenerationMixin
+        return self.layers
+
+    @property
+    def ln_f(self) -> nn.Module:  # For compatibility with RemoteGenerationMixin
+        return self.norm
```
mryab, line 135: Perhaps I couldn't find it, but where exactly is it used in RemoteGenerationMixin?

src/petals/models/mixtral/model.py
```diff
@@ -0,0 +1,177 @@
...
+            attentions=None,
+        )
+
+    @property
+    def word_embeddings(self) -> nn.Embedding:  # For compatibility with RemoteGenerationMixin
+        return self.embed_tokens
+
+    @property
+    def word_embeddings_layernorm(self) -> nn.Module:  # For compatibility with RemoteGenerationMixin
+        return nn.Identity()
```
mryab, line 127: Same here

src/petals/models/mixtral/model.py
```diff
@@ -0,0 +1,177 @@
+from typing import Optional
+
+import hivemind
```
mryab, line 3: Best to import DHT explicitly

src/petals/server/backend.py
justheuristic: This works if layer is sharded, unless the user manually rewrites petals to specify an incompatible sharding config into TensorParallel. Without that, TensorParallel will use a default (automatic) config where
- only linear layers are sharded
- each shard computes the full attention kernel, relu, etc


If user shards the Mixtral model, it will be using default config and run  @artek0chumak  code correctly. Because each layer shard requires full KV cache and has full number of heads in config.


In other words, the current code will work in all cases except when user goes out of their way to modify TensorParallel config in a way that breaks petals servers, which would require changing run_server code.

src/petals/models/mixtral/block.py
```diff
@@ -0,0 +1,113 @@
...
+)
+from transformers.models.mixtral.modeling_mixtral import MixtralDecoderLayer, MixtralModel
+
+
+class WrappedMixtralBlock(MixtralDecoderLayer):
+    def __init__(self, config: MixtralConfig, layer_idx: int):
+        self._attn_implementation = config._attn_implementation
+        self.sliding_window = config.sliding_window
+        self.layer_idx = layer_idx
+        super().__init__(config, layer_idx)
```
mryab, line 18:
```suggestion
        super().__init__(config, layer_idx)
        self._attn_implementation = config._attn_implementation
        self.sliding_window = config.sliding_window
        self.layer_idx = layer_idx
```

mryab: Init of the parent class is usually the first step, any reason why we can't make it that way?

src/petals/models/mixtral/model.py
```diff
@@ -0,0 +1,169 @@
...
+from petals.client.remote_generation import RemoteGenerationMixin, RemotePastKeyValues
+from petals.client.remote_sequential import RemoteSequential
+from petals.models.mixtral.config import DistributedMixtralConfig
+from petals.utils.auto_config import DefaultRevisionMixin
+
+logger = get_logger(__name__)
+
+
+class DistributedMixtralModel(DefaultRevisionMixin, FromPretrainedMixin, PTuneMixin, MixtralModel):
+    """MixtalModel, but all transformer layers are hosted by the swarm"""
```
mryab, line 27:
```suggestion
    """MixtralModel, but all transformer layers are hosted by the swarm"""
```

src/petals/models/mixtral/block.py
artek0chumak: No reason, I've changed the order to the common way.

mryab (approved):

PR #567, opened 2024-03-28
WIP GCP TPU XLA Build

jmikedupont2:
This is the first in a series of prs for getting petals running on v3 tpu on gcp using guix.


src/petals/server/backend.py
```diff
@@ -191,9 +191,9 @@ def shutdown(self):
 
         # Explicitly free the GPU memory. This is not necessary at the time this code is written,
         # but may help to avoid future issues when the module is not garbage-collected for some reasons
-        dummy = torch.tensor([])
-        for p in self.module.parameters():
-            p.data = dummy
+#        dummy = torch.tensor([])
```
ayan4m1, line 194: If this code is broken in a TPU environment, then there should be some runtime check as to whether or not the TPU is being used, the code shouldn't just be commented out entirely imo.

Issue #569, opened 2024-04-08
Error trying to raise Mixtral private swarm server


Qessia:
### Reproduce:
`python3 -m petals.cli.run_server mistralai/Mixtral-8x7B-v0.1 --new_swarm`
or
`python3 -m petals.cli.run_server SanjiWatsuki/TinyMixtral-32x248M --new_swarm`

### Got:
`TypeError: WrappedMixtralBlock.__init__() missing 1 required positional argument: layer_idx`

### System:
- Python 3.10.13
- Cuda 12.2
- torch 2.2.1, torchaudio 2.2.1, torchvision 0.17.1

artek0chumak:
Hello! Thank you for reporting!
We will quickly resolve this issue.

mprishchepo:
Hello!

I observe the same problem. I have tried to diagnose the issue a bit by myselve.

As I understood (if you haven't found it already) the problem is in calculating block size (its parameters). The layer_idx mentioned above is used in load_pretrained_block, but it is not used when calculating block_size and when calculating rps in throughput. 

Very much waiting for a solution.

artek0chumak:
We resolved this issue in recent master update. Just pull new updates. 
Thank tou for noticing the issue and waiting fixes.

Qessia:
Thank you for your quick response!

mprishchepo:
Hi!
Original error of this issue doesn't appear anymore, but I've got another error when I try launching private swarm with Mixtral (with GPU, CPU is ok). Also it doesn't appear when I do the same with StableBeluga2
## System:
- Python3.10
- Torch2.2.2
- Cuda 12.3
- Ubuntu 22.04
## Reproduce
```bash
python3 -m petals.cli.run_server SanjiWatsuki/TinyMixtral-32x248M --new_swarm
```
## Error
```
File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/qessia/.local/lib/python3.10/site-packages/petals/cli/run_server.py", line 235, in <module>
    main()
  File "/home/qessia/.local/lib/python3.10/site-packages/petals/cli/run_server.py", line 219, in main
    server = Server(
  File "/home/qessia/.local/lib/python3.10/site-packages/petals/server/server.py", line 237, in __init__
    throughput_info = get_server_throughput(
[...]

PR #570, opened 2024-04-08
Fix Mixtral-related issues

artek0chumak:
This PR fixes problems related to #569:
- block initialization
- throughput calculation and cache usage
- mixtral in tests

BS is removed for Mixtral and Llama for now. Those models use DynamicCache, which requires special function to change: (see https://github.com/huggingface/transformers/blob/main/src/transformers/cache_utils.py#L161)

### DIFF
--- a/src/petals/server/block_utils.py
+++ b/src/petals/server/block_utils.py
@@ -53,7 +53,7 @@ def get_block_size(
     return round(n_params * bytes_per_value * (1 + eps))
 
 
-def get_model_block(config, **kwargs):
+def get_model_block(config, layer_idx: int = 0):
     """
     The function to create a model block based on the block class
     kwargs argument **only** is necessary for specific classes, like Mixtral.
@@ -61,5 +61,5 @@ def get_model_block(config, **kwargs):
     """
     if config.block_class == WrappedMixtralBlock:
         config = PreTrainedModel._autoset_attn_implementation(config)
-        return config.block_class(config, kwargs.get("layer_idx", 0))
+        return config.block_class(config, layer_idx)
     return config.block_class(config)
--- a/src/petals/server/block_utils.py
+++ b/src/petals/server/block_utils.py
@@ -2,9 +2,10 @@ from typing import Optional, Union
 
 import torch
 from accelerate import init_empty_weights
-from transformers import PretrainedConfig
+from transformers import PretrainedConfig, PreTrainedModel
 
-from petals.utils.convert_block import QuantType, get_model_block
+from petals.models.mixtral.block import WrappedMixtralBlock
+from petals.utils.convert_block import QuantType
 from petals.utils.misc import get_size_in_bytes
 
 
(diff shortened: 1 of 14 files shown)


src/petals/server/block_utils.py
```diff
@@ -50,3 +51,15 @@ def get_block_size(
...
+
+def get_model_block(config, **kwargs):
+    """
+    The function to create a model block based on the block class
+    kwargs argument **only** is necessary for specific classes, like Mixtral.
+    They will not be passed to other block constructors.
+    """
+    if config.block_class == WrappedMixtralBlock:
+        config = PreTrainedModel._autoset_attn_implementation(config)
+        return config.block_class(config, kwargs.get("layer_idx", 0))
```
mryab, line 64: Instead of kwargs, can you change this into one argument to make sure other arguments aren't passed and ignored by mistake?

src/petals/server/throughput.py
```diff
@@ -201,18 +202,25 @@ def measure_compute_rps(
...
         elapsed = 0
         dummy_input = torch.randn(1, n_tokens, config.hidden_size, device=device, dtype=dtype)
-        _, cache = block.forward(dummy_input, use_cache=True)  # Skip the 1st step to exclude the initialization time
+
+        def step(cache_):
+            outputs = block.forward(dummy_input, use_cache=inference, layer_past=cache_ if inference else None)
+            return outputs[1] if inference else None
+
+        cache = step(cache)
+        # Skip the 1st step to exclude the initialization time
```
mryab, line 218: This comment needs to be before L213

tests/test_full_model.py
```diff
@@ -141,6 +141,9 @@ def test_sampling(tokenizer, model, ref_model, max_new_tokens=10):
                 ), f"Sampling is not identical to HF with {options=}, {multiple_calls=}, {inputs.shape=}"
 
 
+@pytest.mark.skipif(
+    MODEL_NAME.lower().find("mixtral"), reason="Mixtral use DynamicCache, that can change based on BS choices"
```
mryab, line 145:
```suggestion
    MODEL_NAME.lower().find("mixtral"), reason="Mixtral uses DynamicCache, which can change based on beam search choices"
```

src/petals/server/block_utils.py
artek0chumak: Yes, done

src/petals/server/throughput.py
artek0chumak: Moved

tests/test_full_model.py
```diff
@@ -141,6 +141,10 @@ def test_sampling(tokenizer, model, ref_model, max_new_tokens=10):
                 ), f"Sampling is not identical to HF with {options=}, {multiple_calls=}, {inputs.shape=}"
 
 
+@pytest.mark.skipif(
+    MODEL_NAME.lower().find("bloom") == -1,
+    reason="Mixtral and Llama uses DynamicCache, which can change based on beam search choices",
```
mryab, line 146:
```suggestion
    reason="Mixtral and Llama use DynamicCache, which can change based on beam search choices",
```

tests/test_full_model.py
```diff
@@ -141,6 +141,10 @@ def test_sampling(tokenizer, model, ref_model, max_new_tokens=10):
                 ), f"Sampling is not identical to HF with {options=}, {multiple_calls=}, {inputs.shape=}"
 
 
+@pytest.mark.skipif(
+    MODEL_NAME.lower().find("bloom") == -1,
```
mryab, line 145:
```suggestion
    "bloom" not in MODEL_NAME.lower(),
```

tests/test_full_model.py
artek0chumak: Done

tests/test_full_model.py
artek0chumak: Done

PR #574, opened 2024-04-11
Fix dummy cache allocation

artek0chumak:
Fix allocation of the dummy key_value cache. It's not used in actual computations, but torch checkers require them to be on the correct device.

### DIFF
--- a/src/petals/server/throughput.py
+++ b/src/petals/server/throughput.py
@@ -206,7 +206,7 @@ def measure_compute_rps(
         block = block.to(dtype)
         block = convert_block(block, 0, config, tensor_parallel_devices, device, quant_type=quant_type, freeze=True)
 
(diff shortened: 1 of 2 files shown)


src/petals/server/throughput.py
```diff
@@ -206,7 +206,7 @@ def measure_compute_rps(
         block = block.to(dtype)
         block = convert_block(block, 0, config, tensor_parallel_devices, device, quant_type=quant_type, freeze=True)
 
-        cache = (DUMMY_KEY_PAST.to(dtype), DUMMY_KEY_PAST.to(dtype))
+        cache = (DUMMY_KEY_PAST.to(dtype).to(device), DUMMY_KEY_PAST.to(dtype).to(device))
```
mryab, line 209: Let's replace chained calls by a single call to `.to`

justheuristic (approved):
LGTM, sorry i took so long

PR #588, opened 2024-07-02
[DRAFT] inv

xtinkt:


src/petals/client/inference_session.py
```diff
@@ -84,7 +84,8 @@ async def _read_inputs_from_queue(queue: asyncio.Queue, input_timeout: Optional[
                 break  # this message means "done sending"
 
     def step(
-        self, inputs: torch.Tensor, prompts: torch.Tensor, hypo_ids: torch.LongTensor, *, step_id: str
+        self, inputs: torch.Tensor, prompts: torch.Tensor, hypo_ids: torch.LongTensor, *,
+        step_id: str, last_validated_position: int
```
justheuristic, line 88: General suggestion: the position can be used not just for speculation, but also for other use cases, such as "regenerate response" option in chat.

Can you please rename last_validated_position (pr-wide) into something neutral? For instance, `start_position`, as in "the index of the first item on inputs in history"? Or anything else you find preferable.

If you insist on `last_validated_position`, please explain your motivation and ignore my comment.

src/petals/client/inference_session.py
```diff
@@ -94,6 +95,12 @@ def step(
         if self.closed:
             raise Exception("Session is closed, cannot perform step")
 
+        if last_validated_position is not None:
+            assert last_validated_position <= self._position
+            self._position = last_validated_position
+            if self.history is not None and self.history.shape[1] >= last_validated_position:
+                self.history = self.history[:, :last_validated_position, :] if last_validated_position > 0 else None
```
justheuristic, line 102: Another rationale: technically, `last_validated_position` is not the index (0-based) of last validated position - it is higher by 1. Otherwise, history[:, :last_validated_position] would delete the last validated token from history. If i got it wrong, please explain.

justheuristic (approved):
LGTM, we can merge this now and begin working on a client-side interface

PR #598, opened 2024-07-23
Added primitives for speculative decoding and tests

xtinkt:


src/petals/models/llama/model.py
```diff
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Union
```
justheuristic, line 1:
```suggestion
from typing import Optional
```

PR #626, opened 2025-08-17
Updating tutorial

jtzl:
I recently got Petals running and want to commit changes in line with what worked for me. I believe there are a couple other updates I plan to submit after this.


jtzl:
Sorry - I just read why Llama is better for a demo.

